PR #11, opened 2013-11-25
First step toward internationaization...

lfourrier:
... here is a french version of the tutorial.


aardappel:
@lfourrier in the latest version, TreeSheets now automatically loads up your translation on any french users system!

Also, I've just made it possible to also translate the UI text if you're interested. Some details are in the "translations" folder. Only tested on Windows so far, but should be tested on other systems soon.

PR #25, opened 2014-12-24
Allow the user to delete empty columns

ormris:
If a selection contains only empty cells, Treesheets will always delete the rows in that selection. These changes allow me to delete empty columns if the selection is thin and empty rows if the selection is wide. 

Comments welcome.


ormris:
Thanks!

PR #29, opened 2015-03-16
Create files for BSD and GNU make with changes inside of Makefile.common

senderghost:
Hello, Wouter van Oortmerssen.

Thanks for addition of Makefile, which I created a while ago on FreeBSD Bugzilla.

My initial version was compatible with BSD and GNU make variants. But you introduced some changes, which used as a GNU extensions and not compatible with BSD make.

I propose to create BSDmakefile and GNUmakefile with changes inside of Makefile.common.
The BSDmakefile (GNUmakefile) is alternative Makefile's name for BSD (GNU) make. This way, they might be used on BSD and Linux operating systems without additional dependencies on particular make system.

Also, I propose to move installation from "$(APPNAME)" target (goal) to "install" and create "deinstall" target for deinstallation. The C++ "-O3" compiler option (for CXXFLAGS) might be used on "release" target. This is useful in case of system-wide configuration (e.g. inside of /etc/make.conf) or custom options and adds more granularity.

The examples of basic usage:
Build executable with system-wide configuraion:
% make
Build executable with custom options from environment:
% env CXX=clang++ CXXFLAGS=-O2 WX_CONFIG=wxgtk2u-3.0-config make
Build and install executable file:
% make install
Deinstall executable file:
% make deinstall
Remove intermediate files:
% make clean
Build release version (with C++ "-O3" compiler option):
% make release
Build debug version (with C++ "-g" compiler option):
% make debug
Also possible to combine targets:
Build release version and install executable file:
% make release install
Build debug version and install executable file:
% make debug install
Deinstall executable and remove intermediate files, if needed:
% make deinstall clean

Tested on FreeBSD 10 for make and gmake variants.

Edit: Added example for environment.

### DIFF
--- a/src/BSDmakefile
+++ b/src/BSDmakefile
@@ -1,0 +1,7 @@
+.if make(debug)
+CXXFLAGS+=	-g
+.elif make(release)
+CXXFLAGS+=	-O3
+.endif
+
+.include "Makefile.common"
--- a/src/GNUmakefile
+++ b/src/GNUmakefile
@@ -1,0 +1,7 @@
+ifeq ($(findstring debug, $(MAKECMDGOALS)), debug)
+CXXFLAGS+=	-g
+else ifeq ($(findstring release, $(MAKECMDGOALS)), release)
+CXXFLAGS+=	-O3
+endif
+
+include Makefile.common
--- a/src/Makefile.common
+++ b/src/Makefile.common
@@ -1,5 +1,4 @@
-# Created by: lightside <lightside@gmx.com>
-# This file released into the public domain
+# Common Makefile for BSD and GNU make
 
 WX_CONFIG?=	wx-config
 WX_CXXFLAGS=	`$(WX_CONFIG) --cxxflags`
@@ -12,22 +11,25 @@ APPNAME=	treesheets
 SRCS=	main.cpp
 OBJS=	$(SRCS:.cpp=.o)
 
-all: CXXFLAGS+= -O3
-all: $(SRCS) $(APPNAME)
+debug: all
 
-debug: CXXFLAGS+= -g
-debug: $(SRCS) $(APPNAME)
+release: all
+
+all: $(SRCS) $(APPNAME)
 
 $(APPNAME): $(OBJS)
 	$(CXX) $(OBJS) $(LDFLAGS) -o $@
-	mv treesheets ../TS
-	rm -f *.o
 
 .o:
 	$(CXX) $(CXXFLAGS) $< -o $@
+
+install: all
+	cp -f $(APPNAME) ../TS
+
+deinstall:
+	rm -f ../TS/$(APPNAME)
 
 clean:
 	rm -f $(APPNAME) *.o
 
 .PHONY: all
-


senderghost:
Honestly to say, it's ok to use GNU make (gmake), if available. It's not hard to install it on FreeBSD from ports, because many software uses it to build. Also, it's easy to convert current Makefile to BSD variant with using sed command:
% sed -i '' -e '/: CXXFLAGS/d' src/Makefile

For this case, I created [makefile](https://github.com/senderghost/treesheets/commits/makefile) branch with related [changes](https://github.com/senderghost/treesheets/commit/073d3e0). If you want, I could create pull request from it.

aardappel:
Yes, a single makefile would be preferable, trying to keep this small, simple, and easy to read :)
Also would prefer to be able to build/install/clean in one go, like my current version.
Thanks for your help though!

senderghost:
Ok, I understood.

The possible problem here is naming of the targets. For example, there is a possibility to change specified targets, but by default they named as follows in FreeBSD ports:
[ALL_TARGET](https://github.com/freebsd/freebsd-ports/blob/4bb1257/Mk/bsd.port.mk#L837) - Default target for sub-make in build stage. Default: all
[INSTALL_TARGET](https://github.com/freebsd/freebsd-ports/blob/4bb1257/Mk/bsd.port.mk#L868) - Default target for sub-make in install stage. Default: install

In case of [deskutils/treesheets](http://www.freshports.org/deskutils/treesheets/) port, I used custom [do-install](https://github.com/freebsd/freebsd-ports/blob/61c7651/deskutils/treesheets/Makefile#L58) target, which installs TreeSheets as shared application on the system. Therefore, the INSTALL_TARGET ignored in this [case](https://github.com/freebsd/freebsd-ports/blob/4bb1257/Mk/bsd.port.mk#L3528).

Though, what you wanted is possible (i.e. run the make command, which does "release", "install" and "clean" targets in order).
I submitted possible [changes](https://github.com/senderghost/treesheets/commit/2ea6142) to [makefile](https://github.com/senderghost/treesheets/commits/makefile) branch.

I guess, the correct way might be to close this pull request from my side and create new one, based on different branch. But I looking forward for your comments, before I do this.

Thanks for attention.

senderghost:
I submitted new [changes](https://github.com/senderghost/treesheets/commit/9d4d86e) to [makefile](https://github.com/senderghost/treesheets/commits/makefile) branch. Its Makefile does as [original](https://github.com/aardappel/treesheets/blob/7950f77/src/Makefile):
make - builds and installs release version
make debug - builds and installs debug version

At least, it might be useful for other maintainers in this state, as well as for you.

Also, what do you think about targets, which create(s) the distribution archive(s) (i.e. archives TS directory with installed treesheets executable)?

I added mentioned changes [there](https://github.com/senderghost/treesheets/commit/1c65642). Some examples:
Create archive for release version:
% make dist-release
Create archive for debug version:
% make dist-debug
Create archive for version with general or custom options:
% make dist-all
Remove distribution archive:
% make dist-clean
Create distribution archive for release version with "treesheets_linux.tar.gz" name, where the type of archive determined automatically, based on supported file extension (i.e. tar.gz, tar.bz2, tar.xz, zip, etc.):
% env DISTFILE=treesheets_linux.tar.gz make dist-release

aardappel:
I'm not sure if I understand the naming issue.. but I am fine with them named differently if that helps on FreeBSD.  I just prefer a single makefile that is simple, for ease of use and future changes. Can you squash that into a single commit PR?

Yes, a dist-release target would be useful, currently it is: tar czvf treesheets_linux64.tar.gz TS
Don't think we need a debug version, nor different archivers though. Naming should default to what I use on treesheets.com. Lets keep it simple :)

senderghost:
Ok, I managed to submit changes to the [single/makefile](https://github.com/senderghost/treesheets/commits/single/makefile) branch with single [commit](https://github.com/senderghost/treesheets/commit/b3aa97e).

I made it simple (from my point of view), configurable from environment and complete.

Also I added automatic determination of DISTFILE (with possibility to change it from environment) based on "uname" output. For example, it's "treesheets_FreeBSD_amd64.tar.gz" on FreeBSD amd64.

aardappel:
Thanks, that works perfectly!

Cherry-picked in https://github.com/aardappel/treesheets/commit/8f0b8b9fe236c4fc2561002235831afd3cbf7593

senderghost:
Thank you too.

PR #30, opened 2015-04-09
Fix "Select First Child" keyboard shortcut

spinda:
The keyboard shortcut for Edit→Selection→Select First Child (Shift+Return) wasn't working (it doesn't look like it was implemented?). This tiny change fixes it.


aardappel:
That's another platform incompatibility. On Windows it already worked before this PR because it goes through OnMenu() instead of Key().. was this on Linux or Mac?

spinda:
This was on Linux.

PR #31, opened 2015-04-09
Minor: Escape ampersand in Edit menu

spinda:
The ampersand in "Layout & Render Style" wasn't properly escaped, so wxWidgets interpreted "& " as a mnemonic and rendered it as "Layout _Render Style".


aardappel:
Thanks!

PR #33, opened 2015-07-01
build tweaks for more flexibility

infinity0:
see commit messages. this helps distros package this application a bit more easily.


aardappel:
Note that TreeSheets isn't exclusive to Linux, we build using Visual Studio and Xcode on other platforms, where this will cause a compilation error. Maybe for now put an #ifndef in there that equates it with __DATE__ instead?

infinity0:
Sure, I have just updated the commits with that.

aardappel:
Thanks!

PR #55, opened 2016-01-28
Implemented the correct handling of non-latin symbols particulary in isalphanum checks

mkeyran:
The wxIsalnum function (being just a wrapper of the standard iswalnum function) and some others are locale-dependent, thus they are working incorrectly when checking non-ASCII characters, since by default LC_ALL is set to C. I added the call of the setlocale function which set the locale to the system environment locale. This fixed navigation by words, opening files with names containing non-ASCII symbols and probably something else.

### DIFF
--- a/src/myapp.h
+++ b/src/myapp.h
@@ -27,6 +27,9 @@ struct MyApp : wxApp
         wxDisableAsserts();
         #endif
 
+        // set locale for the correct handling of non-latin symbols
+        std::setlocale(LC_ALL,"");
+
         bool portable = false;
         wxString filename;
         for (int i = 1; i < argc; i++)
--- a/src/myapp.h
+++ b/src/myapp.h
@@ -28,7 +28,7 @@ struct MyApp : wxApp
         #endif
 
         // set locale for the correct handling of non-latin symbols
-        std::setlocale(LC_ALL,"");
+        std::setlocale(LC_CTYPE,"");
 
         bool portable = false;
         wxString filename;
--- a/src/stdafx.h
+++ b/src/stdafx.h
@@ -57,6 +57,8 @@
 
 #include <wx/display.h>
 
+#include <clocale>
+
 //#include <wx/uiaction.h>
 
 //#include <wx/overlay.h>


aardappel:
Thanks!

Might it not be safer to use `LC_CTYPE` ? For example, there may be code in there that relies on numeric formatting using `.` for fractions, which this would also change.

As for `LC_CTYPE`, can you give me an example of a character that will be classified differently? I tested the word splitting with various unicode characters and it seemed to work.

mkeyran:
Yes, you are right, `LC_TYPE` solves the problems. It seems that any of the cyrillic characters are misclassified, for example `Б`.  This is caused by the fact that for some reasons the default locale in the running program is 'C', which can be seen from 
 
       wxLogMessage(std::setlocale(LC_CTYPE, nullptr));
       wxLogMessage(std::setlocale(LC_ALL, nullptr));

Here is my `locale` output:

    LANG=ru_RU.utf8
    LC_CTYPE=ru_RU.utf8
    LC_NUMERIC="ru_RU.utf8"
    LC_TIME="ru_RU.utf8"
    LC_COLLATE="ru_RU.utf8"
    LC_MONETARY="ru_RU.utf8"
    LC_MESSAGES="ru_RU.utf8"
    LC_PAPER="ru_RU.utf8"
    LC_NAME="ru_RU.utf8"
    LC_ADDRESS="ru_RU.utf8"
    LC_TELEPHONE="ru_RU.utf8"
    LC_MEASUREMENT="ru_RU.utf8"
    LC_IDENTIFICATION="ru_RU.utf8"
    LC_ALL=

Oh, and I forgot to mention that this behaviour is seen in Linux.

mkeyran:
By the way, it seem that exactly the same thing is done in QtCoreApplication in Qt for the same reason.
http://stackoverflow.com/questions/25661295/why-does-qcoreapplication-call-setlocalelc-all-by-default-on-unix-linux .

aardappel:
Ok, we'll see if this improves things.

Still pretty weird. You'd think that `iswalnum` operating on unicode characters could simply do the right thing for all languages at once, regardless of locale. I don't see why, say, russian characters should be treated any differently if my locale is english vs russian.

PR #57, opened 2016-02-20
Removed dependence on c11 features (auto type)

vo1stv:
Not sure what your motivation for using c11 features is.
I believe less is more; the changes in this branch allow the code to be built without c11 features.
I have built this on Ubuntu, but I haven't tested on other flavours of Linux or on Windows or Mac, so if your motivation for c11 has to do with cross-platform compatibility, you can reject this pull request.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -6,7 +6,7 @@ WX_CXXFLAGS=	`$(WX_CONFIG) --cxxflags`
 WX_LIBS=	`$(WX_CONFIG) --libs aui adv core xml net`
 PACKAGE_VERSION?=	__DATE__
 CXX?=	c++
-CXXFLAGS+=	-DPACKAGE_VERSION=$(PACKAGE_VERSION) $(WX_CXXFLAGS)
+CXXFLAGS+=	-DPACKAGE_VERSION=$(PACKAGE_VERSION) -std=c++11 $(WX_CXXFLAGS)
 LDFLAGS+=	$(WX_LIBS)
 
 ARCH=	$(firstword $(shell uname -m))
--- a/src/Makefile
+++ b/src/Makefile
@@ -6,7 +6,7 @@ WX_CXXFLAGS=	`$(WX_CONFIG) --cxxflags`
 WX_LIBS=	`$(WX_CONFIG) --libs aui adv core xml net`
 PACKAGE_VERSION?=	__DATE__
 CXX?=	c++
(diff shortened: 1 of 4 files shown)


dahn-d:
Why and how may `auto` hurt?

aardappel:
I personally prefer code that uses `auto` everywhere, it is more readable to me, less type clutter. But this code mostly predates my switch to C++11, so I don't mind keeping it compatible with older C++ for now. Just a few corrections.

src/Makefile
```diff
@@ -6,7 +6,7 @@ WX_CXXFLAGS=	`$(WX_CONFIG) --cxxflags`
 WX_LIBS=	`$(WX_CONFIG) --libs aui adv core xml net`
 PACKAGE_VERSION?=	__DATE__
 CXX?=	c++
-CXXFLAGS+=	-DPACKAGE_VERSION=$(PACKAGE_VERSION) -std=c++11 $(WX_CXXFLAGS)
```
aardappel: Let's keep this in. TreeSheets is primarily developed on Windows, and I don't want my linux builds to break when a commit accidentally uses auto.

src/document.h
```diff
@@ -1225,18 +1225,18 @@ struct Document
             case A_CUSTKEY:
             {
                 wxArrayString strs;
-                auto &ms = sys->frame->menustrings;
-                for (auto it = ms.begin(); it != ms.end(); ++it) strs.push_back(it->first);
+                MyFrame::MenuString & ms = sys->frame->menustrings;
```
aardappel: Please use the existing style of the surrounding code, meaning no space after &

src/myframe.h
```diff
@@ -1,5 +1,7 @@
 struct MyFrame : wxFrame
 {
+    typedef std::vector<std::pair<wxString, wxString> > MenuString;
+    typedef std::vector<std::pair<wxString, wxString> >::iterator MenuStringIterator;
```
aardappel: This is why you use auto :)

vo1stv: Agreed auto works nicely for iterators.

src/Makefile
vo1stv: Not sure what you mean here.

vo1stv:
I'm a bit new to the GitHub PR workflow.  I think I've addressed your concerns (waiting on clarification for Makefile note)

aardappel:
Looks good now, thanks.

PR #58, opened 2016-03-20
Remove unused rule

infinity0:
The rule `.o` matches nothing, and treesheets was relying on make's implicit pre-defined `*.o` rule anyway. So just delete it to avoid confusing people.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -32,9 +32,6 @@ all: $(SRCS) $(APPNAME)
 $(APPNAME): $(OBJS)
 	$(CXX) $(OBJS) $(LDFLAGS) -o $@
 
-.o:
-	$(CXX) $(CXXFLAGS) $< -o $@
-
 install: all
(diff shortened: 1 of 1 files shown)


aardappel:
Does the implicit rule have the same definition as the one given? Is it worth making it into `*.o` such that if someone wants to change this rule in the future it is easy (as opposed to having to add it) ?

infinity0:
https://www.gnu.org/software/make/manual/html_node/Catalogue-of-Rules.html "... with a recipe of the form ‘`$(CXX) $(CPPFLAGS) $(CXXFLAGS) -c`’. i.e. it additionally honours `CPPFLAGS` which is flags for the C pre-processor.

I don't think it's worth keeping - I don't foresee a use-case that couldn't be satisfied with the existing variables.

aardappel:
Ok, makes sense, thanks!

PR #61, opened 2016-05-13
Update treesheets.desktop

infinity0:
Add categories, keywords, mimetype, and a Terminal=false for completeness


aardappel:
Thanks!

PR #66, opened 2016-05-18
Add icons in standard formats and drop deprecated formats

infinity0:
https://standards.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html

says that only PNG, XPM and SVG are supported, and indeed neither the BMP or
GIF files work when I install them to the standard system locations.

The PNG files were generated deterministically from the ICO file like this:

$ convert icon1.ico treesheets.png && strip-nondeterminism treesheets-[01].png
$ sha256sum treesheets*.png
061597a2e5887576b8d77443406d6027847ec4402d060405a59dd00ca29eeb1b  treesheets-0.png
80bcb38dcb111c2e5dfa29aa3f8ea4f5ea35ca8757e874e1be054c35e5ba43b4  treesheets-1.png

See strip-nondeterminism: https://packages.debian.org/sid/strip-nondeterminism

Strictly, you don't actually need these PNG files, but it's a nice convenience
so your users don't have to install imagemagick themselves to build it.


aardappel:
Thanks!

Though, TreeSheets runs on platforms other than Linux, and I am not sure if the two files you're deleting aren't referred to by any project/installer files on the other 2 platforms.

Also, the new files you created aren't referred to by anything, so what is their use case?

infinity0:
`git grep bmp` and `git grep gif` don't show up anything of interest, so I figured it was fine to delete those.

The PNG files are not referred to by anything else, correct, but neither is `treesheets.desktop`. In the Debian package, I am installing them to specific locations. It would be nice (actually for Debian too) if Treesheet's Makefile had in `install` target that did this, I just would need to spend some time figuring out if the paths I used are platform-independent - or feel free to do this research yourself. But you should only do this only if you're happy to maintain it in the long run. (I'd prefer to keep my role restricted to only Debian-specific things, for the context of this project.)

For reference the paths are

/usr/share/applications/treesheets.desktop
/usr/share/icons/hicolor/16x16/apps/treesheets.png
/usr/share/icons/hicolor/32x32/apps/treesheets.png

and in the Makefile it would look something like `$(DESTDIR)$(datadir)/applications` etc, see https://www.gnu.org/prep/standards/html_node/Directory-Variables.html for details.

aardappel:
Ok, makes sense.
I welcome your debian/linux related improvements, as long as they don't affect the other platforms.

PR #67, opened 2016-05-20
 CI Passes on linux and os/x

vo1stv:
CI Passes on linux and os/x 

Experimental support for wxWidgets 3.1 in linux
To flush cache, edit .travis.yml to set  RMRF_WXWIDGETS to a non-empty string 
(this is the default for osx: for faster CI testing, reset osx to RMRF_WXWIDGETS="")
CI integration occurs only on master and testing branch


aardappel:
Thanks!

PR #81, opened 2017-07-03
Add CMake build.

tamaskenez:
Thanks for this great tool!
For me it was easier to build with CMake so I added this script. Feel free to accept or reject.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,40 @@
+# Sample build with makefile generator
+#
+#     cd <into-treesheets-source>
+#     cmake -H. -Bb -DCMAKE_BUILD_TYPE=Release \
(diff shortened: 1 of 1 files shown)


aardappel:
Thanks!
Not sure how I feel about relying on the system wxWidgets, since it is often out of date, and of course Windows has no concept of it being installed.
But for now, this doesn't hurt to have also.

tamaskenez:
It's not tied to system wxWidgets, even on Windows the user can build the latest one from sources, install it to a local dir, tell CMake (the FindwxWidgets.cmake module) about that dir and done.

In the meantime I've just tested on ub14, works fine with system wxwidgets 3.0.

aardappel:
Cool, thanks :)

ngirard:
I confirm it's a valuable addition, and it works fine here on Ubuntu 17.04 !

PR #88, opened 2017-08-22
With WxWidgets, Unicode seems to be the default and there appears to …

owencook:
…be no optimization option


aardappel:
I'm not sure what you're trying to accomplish here.. you think we should turn off unicode and optimization?

aardappel:
Ok, I see.
There are now 2 binary files in your commit, can you remove them?
You can also remove the description of the missing flags from the readme. I only wanted to know why they were removed here on this PR.

owencook:
Apologies for the binaries, they are now removed.  GIT ignorance on my part.

aardappel:
Ok, thanks!

Issue #95, opened 2018-01-29
idea: shift+enter inserts new child grid when none exists


gussmith23:
Using shift+enter is a natural shortcut to selecting a cell's child grid. If a grid doesn't exist, I think it would be equally natural to have shift+enter create the child grid. Currently, without an INS key, inserting has to be done with the mouse. This would (possibly, depending on the application's design) be an easy way to resolve this complaint without implementing a key mapping system.

aardappel:
I agree that would be a natural extension of the existing behavior.

gussmith23:
By the way -- I intend to take a crack at some of these bugs I'm opening, seeing as I've opened a number of them now, haha. It may have to wait until the summer, when I'm more free. I'm thinking this might be my first one to take on (if someone doesn't take it first!)

aardappel:
@gussmith23 that be appreciated :)

aardappel:
Fixed by: https://github.com/aardappel/treesheets/pull/97

gussmith23:
beautiful! thanks all!

PR #97, opened 2018-02-07
Make 'Shift + Enter' create a new grid in a cell

AveryGlitch:
If there isn't a grid in a cell, shift enter should make one (as mentioned in [this issue](https://github.com/aardappel/treesheets/issues/95)

The change itself is very simple - instead of the code calling NoGrid() if there's no grid, it re-calls Action as if Insert had been pressed.

I'd like to also modify the tutorial (cell 10) to say "INSERT or SHIFT + ENTER", but I don't know where I would submit that change.

### DIFF
--- a/src/document.h
+++ b/src/document.h
@@ -1486,7 +1486,7 @@ struct Document {
             }
 
             case A_ENTERGRID:
-                if (!c->grid) return NoGrid();
+                if (!c->grid) Action(dc, A_NEWGRID);
                 selected = Selection(c->grid, 0, 0, 1, 1);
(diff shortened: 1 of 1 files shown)


aardappel:
Hah, an elegant change, thanks :)

What do you mean "where I would submit that change"? What's wrong with changing the file and submitting a PR?

AveryGlitch:
sorry I meant "I don't know where the tutorial file is"

aardappel:
It's in `TS/examples/tutorial.cts`.

PR #132, opened 2019-08-22
Fix src/Makefile and sort list of files in CMakeLists.txt

senderghost:
Hello, Wouter van Oortmerssen.

I propose to fix src/Makefile and sort list of files in CMakeLists.txt.


aardappel:
Thanks!

PR #137, opened 2019-09-30
Include wxWidgets version in the about dialog

learn-more:
Also include some small readme tweaks for building with Visual Studio


aardappel:
Thanks :)

Issue #106, opened 2018-08-14
ctrl + delete to delete whole words


ogrotten:
Related to #99, except that it doesn't insert a character while not deleting the word.

aardappel:
Fixed by: https://github.com/aardappel/treesheets/pull/138

Issue #99, opened 2018-02-19
ctrl+backspace to delete whole words 


gussmith23:
If this is currently a feature, it doesn't work on Windows. It inserts this character: [""](https://www.fileformat.info/info/unicode/char/007f/index.htm) (DELETE character; I guess that makes sense!)

ctrl+backspace seems to be a pretty standard combo in other editors -- hoping it can be added here too!

aardappel:
Yup, looks like that isn't implemented, and would make sense to be added.

aardappel:
Fixed by: https://github.com/aardappel/treesheets/pull/138

PR #138, opened 2019-10-02
Add options to delete the word before and after the cursor

learn-more:
The hotkeys ctrl+backspace and ctrl+delete can be used
#99
#106

### DIFF
--- a/src/document.h
+++ b/src/document.h
@@ -1126,6 +1126,16 @@ struct Document {
                 ZoomOutIfNoGrid(dc);
                 return nullptr;
 
+            case A_DELETE_WORD:
+                if (c && selected.TextEdit()) {
+                    if (selected.cursor == c->text.t.Len()) return nullptr;
+                    c->AddUndo(this);
+                    c->text.DeleteWord(selected);
+                    Refresh();
+                } 
+                ZoomOutIfNoGrid(dc);
+                return nullptr;
+
             case A_COPYCT:
             case A_CUT:
             case A_COPY:
@@ -1555,6 +1565,14 @@ struct Document {
                 else
                     Refresh();
                 selected.ExitEdit(this);
+                return nullptr;
+
+            case A_BACKSPACE_WORD:
+                if (selected.cursorend == 0) return nullptr;
+                c->AddUndo(this);
+                c->text.BackspaceWord(selected);
+                Refresh();
+                ZoomOutIfNoGrid(dc);
                 return nullptr;
 
             // FIXME: this functionality is really SCHOME, SHOME should be within line
--- a/src/myframe.h
+++ b/src/myframe.h
@@ -381,6 +381,10 @@ struct MyFrame : wxFrame {
             MyAppend(
                 editmenu, A_BACKSPACE, _(L"Delete Before\tBACK"),
                 _(L"Deletes the column of cells before the selected grid line, or the row above"));
+            MyAppend(editmenu, A_DELETE_WORD, _(L"Delete Word After\tCTRL+DEL"),
+                     _(L"Deletes the entire word after the cursor"));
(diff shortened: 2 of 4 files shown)


src/myframe.h
```diff
@@ -381,6 +381,10 @@ struct MyFrame : wxFrame {
             MyAppend(
                 editmenu, A_BACKSPACE, _(L"Delete Before\tBACK"),
                 _(L"Deletes the column of cells before the selected grid line, or the row above"));
+            MyAppend(editmenu, A_DELETE_WORD, _(L"&Delete After\tCTRL+DEL"),
+                     _(L"Deletes the entire word after the cursor"));
+            MyAppend(editmenu, A_BACKSPACE_WORD, _(L"Delete Before\tCTRL+BACK"),
+                     _(L"Deletes the entire word before the cursor"));
```
aardappel: Add `Word` to these 2 menu strings to differentiate them from the non-ctrl version?

src/text.h
```diff
@@ -344,6 +344,14 @@ struct Text {
         while (s.cursorend < (int)t.Len() && wxIsalnum(t[s.cursorend])) s.cursorend++;
     }
 
+    void SelectWordBefore(Selection &s) {
+        if (s.cursor <= 1) return;
+        s.cursorend = s.cursor--;
+        if (!wxIsalnum(t[s.cursor])) return;
+        while (s.cursor > 0 && wxIsalnum(t[s.cursor - 1])) s.cursor--;
+        while (s.cursorend < (int)t.Len() && wxIsalnum(t[s.cursorend])) s.cursorend++;
```
aardappel: These 3 lines are the same as the previous function.. pull them out into a helper function? or add a `bool before` arg? Whatever you think is nicer.

src/text.h
```diff
@@ -387,6 +395,19 @@ struct Text {
                 --s.cursorend;
             };
     }
+    void DeleteWord(Selection &s) {
+        SelectWord(s);
+        if (!RangeSelRemove(s))
+            if (s.cursor < (int)t.Len()) { t.Remove(s.cursor, 1); };
```
aardappel: Call `Delete`? Or call `SelectWord` in the caller, and remove this function entirely?

src/text.h
```diff
@@ -387,6 +395,19 @@ struct Text {
...
+        if (!RangeSelRemove(s))
+            if (s.cursor < (int)t.Len()) { t.Remove(s.cursor, 1); };
+    }
+    void BackspaceWord(Selection &s) {
+        SelectWordBefore(s);
+        if (!RangeSelRemove(s))
+            if (s.cursor > 0) {
+                t.Remove(--s.cursor, 1);
+                --s.cursorend;
+            };
```
aardappel: Same here, call `Backspace` and/or move `SelectWordBefore` to caller.

src/myframe.h
learn-more: Totally missed this, thanks

src/text.h
learn-more: Yeah, I more or less kept to the style of the surrounding code.
The other functions were also roughly the same so I duplicated the code here as well.

aardappel: Well, the style of the surrounding code is to not repeat yourself :)

learn-more:
Much cleaner now, thanks for the feedback!

src/document.h
```diff
@@ -1126,6 +1126,26 @@ struct Document {
                 ZoomOutIfNoGrid(dc);
                 return nullptr;
 
+            case A_BACKSPACE_WORD:
+                if (c && selected.TextEdit()) {
```
aardappel, line 1130: these two actions can be moved to below around line 1569 where it says `if (!selected.TextEdit()) return _(L"only works in cell text mode");`, so this `if` can be omitted.

aardappel:
Thanks!

PR #167, opened 2020-05-24
Update Windows build instructions to use VS2019

aryehof:
Updated the Windows build instructions to use VS2019, so the solution will build correctly without issues (unlike VS2017). Also specified that the wxWidgets builds should be x64 explicitly, rather than Win32 as is the default.


aardappel:
Thanks!

PR #186, opened 2021-07-03
Screen sharing improvements: zen mode, window size API

se3142:
This pull request offers two new features:

- zen mode: hide toolbar and statusbar
- window size scripting API

Intended use case:
When sharing TreeSheets over messaging tools, it may be useful to have a predefined window size with reduced user interface. 

Thanks for checking this request!

### DIFF
--- a/TS/scripts/Set Window Size.lobster
+++ b/TS/scripts/Set Window Size.lobster
@@ -1,0 +1,7 @@
+// Set the window size to a predefined value.
+// Use case: Program sharing over common messaging software
+
+//ts_set_window_size(1920, 1080)
+ts_set_window_size(1366, 768)
+//ts_set_window_size(1280, 720)
+//ts_set_window_size(1024, 768)
--- a/TS/scripts/Set Window Size.lobster
+++ b/TS/scripts/Set Window Size.lobster
@@ -1,7 +1,0 @@
-// Set the window size to a predefined value.
-// Use case: Program sharing over common messaging software
-
-//ts_set_window_size(1920, 1080)
-ts_set_window_size(1366, 768)
-//ts_set_window_size(1280, 720)
-//ts_set_window_size(1024, 768)
--- a/src/myframe.h
+++ b/src/myframe.h
@@ -21,6 +21,7 @@ struct MyFrame : wxFrame {
     bool watcherwaitingforuser;
     double csf, csf_orig;
     std::vector<std::string> scripts_in_menu;
+    bool zenmode;
 
     wxString GetPath(const wxString &relpath) {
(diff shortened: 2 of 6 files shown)


TS/scripts/Set Window Size.lobster
```diff
@@ -0,0 +1,7 @@
+// Set the window size to a predefined value.
+// Use case: Program sharing over common messaging software
+
+//ts_set_window_size(1920, 1080)
+ts_set_window_size(1366, 768)
```
aardappel, line 5: I don't think it makes sense to add an example script file that has these hardcoded sizes in it, as that isn't useful to anyone

src/myframe.h
```diff
@@ -925,6 +928,23 @@ struct MyFrame : wxFrame {
                 ShowFullScreen(!IsFullScreen());
                 if (IsFullScreen()) sw->Status(_(L"Press F11 to exit fullscreen mode."));
                 break;
+            case A_ZEN_MODE:
+                if (!IsFullScreen()) {
+                    wxToolBar *wtb = this->GetToolBar();
+                    wxStatusBar *wsb = this->GetStatusBar();
+                    if (wtb != nullptr)
```
aardappel, line 935: these are `nullptr` when they're currently not shown? if so how do they turn back on?

aardappel (commented):
Thanks, nice functionality!

src/myframe.h
se3142: Depending on the user setting, the toolbar and statusbar remain nullptr. These are initalized on app start:

1. https://github.com/aardappel/treesheets/blob/master/src/myframe.h#L675
2. https://github.com/aardappel/treesheets/blob/master/src/myframe.h#L607

In my understanding - if the user setting defaults to no statusbar, then keep this.
Does this sound right or do I get it wrong?

TS/scripts/Set Window Size.lobster
se3142: Thanks for this feedback. I did not know how to get a user input - is there a way to show a choice dialog or something?

src/myframe.h
aardappel: ah ok, then that makes sense

TS/scripts/Set Window Size.lobster
aardappel: I don't think we do currently... but even so, inputting size in dialog is not that useful compared to just resizing the window. I suggest to remove this file from the PR and keep it for your personal use.

se3142: Agreed! I removed the script file. Is everything ok now?

aardappel:
Thanks!

Issue #139, opened 2019-10-03
Progress to the right cell when editing?


sphh:
When I edit a cell and I press `Enter`, the 'cursor' selects the grid line **below** the current cell. Is there a way to select the grid line **right** of the current cell?

I often tend to fill the cells column by column working my way from left to right. At the moment I have to press `Enter`, `Up` and `Right` to get to the cell right of the current cell …

aardappel:
no key for that yet.. maybe we do can alt-enter or similar for that?

sphh:
I just noticed, that `TAB` and `Shift+TAB` moves the selection to the right and left cell.

sphh:
Sorry, I have to re-open this issue. Jumping to the cell right or left of the current is only working, if there is a cell right or left. If the selected and edited cell is on the very border, `TAB` would not select the grid line.

Maybe instead of using `Alt+Enter` or similar, it might be better to change the behaviour of the `TAB` (and `Shift+Tab` to select the grid line instead of jumping to the next cell. Then the behaviour of the `TAB` key would be very similar to the `Enter` key.

aardappel:
Yeah that should be easy to change..

aardappel:
TAB currently never select grid-lines, it merely steps thru all the cells, and also goes into edit mode. So making an exception to that and make it go to the outer grid line only for the last cell would be a bit inconsistent. In fact, to make that fully useful, it should then stop on the right outer grid line followed by the left one. That would be a bit weird.

So I think if we want the equivalent of enter going to the gridline below also to the right, it should be a different key.

sphh:
You are right, this would be a bit weird.

What I meant, was that `TAB` should behave horizontally like `Enter` does vertically! Or very similar at least: `TAB` should select the grid line right of the current cell or the cell right of the currently selected grid line. So instead of going to the next cell, `TAB` could also select the grid line in between. `Shift+TAB` would go to the previous grid line or cell.

What do you think?

aardappel:
I don't think changing the behavior of TAB would be good for current users that are used to it. They'd need twice as many keystrokes to navigate.

sphh:
Yes, tabbing twice is the draw-back! IMHO the advantages outweigh the disadvantages. Which key do you have in mind?

I just played with the arrow keys. They also show a non-consistent behaviour:

-  Pressing the `Right` key moves to the next grid line or cell. In edit mode and being at the end of the text it does nothing.
[...]

PR #187, opened 2021-07-04
Alt+Enter progress to the right cell

se3142:
This pull request implements Alt+Enter to continue editing to the right.
Related issue and change proposal: https://github.com/aardappel/treesheets/issues/139#issuecomment-544167524

Thanks for checking this pull request.

### DIFF
--- a/src/document.h
+++ b/src/document.h
@@ -764,7 +764,14 @@ struct Document {
             switch (k) {
                 case WXK_BACK:  // no menu shortcut available in wxwidgets
                     return Action(dc, A_BACKSPACE);
-                case WXK_RETURN: return Action(dc, shift ? A_ENTERGRID : A_ENTERCELL);
+                case WXK_RETURN:
+                    if (shift) {
+                       return Action(dc, A_ENTERGRID);
+                    } else if (alt) {
+                        return Action(dc, A_PROGRESSCELL);
+                    } else {
+                        return Action(dc, A_ENTERCELL);
+                    }
                 case WXK_ESCAPE:  // docs say it can be used as a menu accelerator, but it does not
                                   // trigger from there?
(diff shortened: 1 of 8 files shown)


src/document.h
```diff
@@ -1313,6 +1320,17 @@ struct Document {
                 }
                 return nullptr;
 
+            case A_PROGRESSCELL: {
```
aardappel, line 1323: Looks like this case could be merged with the case above it, with only `A_DOWN` and `A_RIGHT` being different?

src/mycanvas.h
```diff
@@ -114,7 +115,8 @@ struct TSCanvas : public wxScrolledWindow {
         // (scrolling) don't work.
         // The 128 makes sure unicode entry on e.g. Polish keyboards still works.
         // (on Linux in particular).
-        if ((ce.GetModifiers() == wxMOD_ALT) && (ce.GetUnicodeKey() < 128)) {
+        // Special case: ALT+RETURN shall be processed.
+        if (ce.GetKeyCode() != WXK_RETURN && (ce.GetModifiers() == wxMOD_ALT) && (ce.GetUnicodeKey() < 128)) {
```
aardappel, line 119: `enter` or `alt + enter` is used in the menus?
For consistency, remove () from the latter 2 conditions, or add it to the first

aardappel (commented):
Nice! thanks for implementing this!

src/document.h
se3142: Agreed - both handlers are merged now.

src/mycanvas.h
se3142: I revisited the wxWidgets event handling and got to know, that shortcuts are auto-assigned by MyAppend. I simplified my implementation.

aardappel:
Thanks!

PR #188, opened 2021-07-05
New shortcuts for program and new operator reference

se3142:
This pull request improves the accessibility of the programming functionality:

- Shortcut for Run (ALT+F5)
- Shortcuts for setting the field type (ALT+? -> D=data, o=operator, a=assign, r=read, .=h-view, ,=v-view)
- New help document: operator reference (ALT+F1)

Thanks for checking this request.


src/myframe.h
```diff
@@ -555,21 +555,22 @@ struct MyFrame : wxFrame {
         }
 
         wxMenu *markmenu = new wxMenu();
-        MyAppend(markmenu, A_MARKDATA, _(L"&Data"));
-        MyAppend(markmenu, A_MARKCODE, _(L"&Operation"));
-        MyAppend(markmenu, A_MARKVARD, _(L"Variable &Assign"));
-        MyAppend(markmenu, A_MARKVARU, _(L"Variable &Read"));
-        MyAppend(markmenu, A_MARKVIEWH, _(L"&Horizontal View"));
-        MyAppend(markmenu, A_MARKVIEWV, _(L"&Vertical View"));
+        MyAppend(markmenu, A_MARKDATA, _(L"&Data\tALT+d"));
```
aardappel, line 558: do these not clash with the menus or anything else?

src/myframe.h
```diff
@@ -555,21 +555,22 @@ struct MyFrame : wxFrame {
...
         wxMenu *langmenu = new wxMenu();
-        MyAppend(langmenu, A_RUN, _(L"&Run"));
+        MyAppend(langmenu, A_RUN, _(L"&Run\tALT+F5"));
         langmenu->AppendSubMenu(markmenu, _(L"&Mark as"));
         MyAppend(langmenu, A_CLRVIEW, _(L"&Clear Views"));
 
         wxMenu *helpmenu = new wxMenu();
         MyAppend(helpmenu, A_ABOUT, _(L"&About..."));
         MyAppend(helpmenu, A_HELPI, _(L"Load interactive &tutorial...\tF1"));
+        MyAppend(helpmenu, A_HELP_OP_REF, _(L"Load operation reference...\tALT+F1"));
```
aardappel, line 573: I guess worth putting these in their own file rather than a section of tutorial?
In their own file is neater, especially if we expand the range of ops.
But in the tutorial people will come across them naturally.
Maybe make sure to add clear reference to the tutorial (where it already has a code example): "for more details on operations, see Help -> Load operation reference (ALT+F1)" or similar.

aardappel (commented):
Thanks for adding these!

src/myframe.h
se3142: I did check this but I revisited your concern. RegEx `\\talt\+[ador.,]"` on the whole project returns 

1. myframe.h:563 MyAppend(markmenu, A_MARKVIEWV, _(L"&Vertical View\tALT+,"));
2. myframe.h:558MyAppend(markmenu, A_MARKDATA, _(L"&Data\tALT+d"));
3. myframe.h:559MyAppend(markmenu, A_MARKCODE, _(L"&Operation\tALT+o"));
4. myframe.h:560MyAppend(markmenu, A_MARKVARD, _(L"Variable &Assign\tALT+a"));
5. myframe.h:561MyAppend(markmenu, A_MARKVARU, _(L"Variable &Read\tALT+r"));
6. myframe.h:562MyAppend(markmenu, A_MARKVIEWH, _(L"&Horizontal View\tALT+."));

There should be no collision.

src/myframe.h
se3142: Good point. I updated all three tutorial files (EN, DE, FR).

src/myframe.h
aardappel: I'm talking about being able to open the menus using the keyboard. For example, currently pressing ALT+e brings up the edit menu, and then ALT+d selects `Delete After` from the menu. Your shortcuts also have a ALT+d. Can you check that the above sequence still works?

Or more directly, if we ever added a Debug menu, we couldn't reach that with ALT+d. The ALT+LOWERCASE space is shared between menus and this specific menu.

If they indeed can clash, we might be better off with CTRL+ALT+.. for these?

se3142: Oh I see. I rarely use this menubar navigation - thanks! In case of possible l10n there is no reliable way to avoid all ALT+? methods. I changed to CTRL+ALT+? pattern.

aardappel: I don't use it either.. but some might :)

aardappel:
Thanks!

aardappel:
Just checked the new operations reference locally, nicely done :)

aardappel:
Your new operator reference made me realize how sloppy the memory management of the evaluator is.. so replaced a bunch of it by `unique_ptr`, so it is a bit clearer who owns what. It's still a mess, but this is a first step towards modernizing it a bit: https://github.com/aardappel/treesheets/commit/5b524a61ca14babd58dc18d941ca9a59a56416c4

PR #189, opened 2021-07-06
Added celltype to XML import / export

se3142:
A small improvement: XML import/export now keeps track of cell types (data, operator, variable, view).
Thanks for reviewing!


aardappel:
Thanks :)

PR #190, opened 2021-07-15
Fleshed out the Linux build instructions.

DDR0:
A few key steps were missing from the Linux build instructions, I believe. It took me a while to figure it out, so I'm going to try to update the instructions with what I've learned.


aardappel:
`-G "Unix Makefiles"` should be default on Linux, what does it say when you leave it out?

And `&& make` is implied, but it doesn't hurt to mention it. I usually say `make -j8` or whatever though, to remind people to specify threads, since `make` doesn't do that by default and it's rather slow without it.

jtojnar:
This should no longer be relevant.

PR #196, opened 2022-01-01
Update ts.po

albanobattistella:


aardappel:
Thanks! Can you run the command in `compile.bat` to update the `ts.mo` file as well?

Issue #133, opened 2019-09-04
"make install" installs things in quite unusual locations


probonopd:
`make install` installs things in quite unusual locations.

When using

```
      cmake . -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/usr
      make -j$(nproc)
      make DESTDIR=appdir -j$(nproc) install ; find appdir/
```

I would expect to get something like

```
appdir/usr/bin/treeshets
appdir/usr/share/treeshets/docs/...
appdir/usr/share/treeshets/examples/...
appdir/usr/share/applications/treeshets.desktop
appdir/usr/share/icons/hicolor/scalable/apps/treesheets.svg
...
```

but this is what actually happens:

```
appdir/
appdir/usr
appdir/usr/docs
appdir/usr/docs/file_format_spec.txt
appdir/usr/docs/history.txt
appdir/usr/docs/tutorial.html
appdir/usr/docs/script_reference.html
appdir/usr/docs/donations.html
appdir/usr/docs/screenshots.html
appdir/usr/docs/images
appdir/usr/docs/images/screenshots
appdir/usr/docs/images/screenshots/screenshot_tutorial.png
appdir/usr/docs/images/screenshots/screenshot_personel.png
appdir/usr/docs/images/screenshots/screenshot_todo_linux.png
appdir/usr/docs/images/screenshots/screenshot_todo.png
appdir/usr/docs/images/screenshots/screenshot_todo_half.png
appdir/usr/docs/images/screenshots/screenshot_tutorial2.png
appdir/usr/docs/images/screenshots/screenshot_sales.png
appdir/usr/docs/images/screenshots/screenshot_todo_mac.png
appdir/usr/docs/images/screenshots/screenshot_unicode.png
appdir/usr/docs/images/excl.png
appdir/usr/docs/images/check.png
appdir/usr/docs/images/treesheets_logo.png
appdir/usr/docs/images/stop.png
appdir/usr/docs/images/documentation
appdir/usr/docs/images/documentation/textsize.png
appdir/usr/docs/images/documentation/edit.png
appdir/usr/docs/images/documentation/gridsel1.png
appdir/usr/docs/images/documentation/pregrid.png
appdir/usr/docs/images/documentation/zoomed.png
appdir/usr/docs/images/documentation/gridsel2.png
appdir/usr/docs/images/documentation/new.png
appdir/usr/docs/images/documentation/gridline.png
[...]

PR #201, opened 2022-03-10
cmake: Allow out-of-tree builds

jtojnar:
Software on most Linux systems is expected to follow [Filesystem Hierarchy Standard](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/index.html. Let’s update the CMake build to support it so that each Linux distro does not have to do this on its own. On Windows, we can still have it relocatable.

- [x] install translations to datadir https://docs.wxwidgets.org/trunk/overview_i18n.html
- [x] install and allow loading rest of the stuff from datadir

Fixes: #133

Eventually, we might want to switch to [Meson](https://mesonbuild.com/) instead of CMake, which might allow us to get rid of the Visual Studio boilerplate since it can generate it.


.gitignore
```diff
@@ -17,14 +17,8 @@ TreeSheets/\.vs/treesheets/v15/
 build/treesheets/language/
 
 TS/scripts/\.con\.log
-TS/treesheets
-TS/2
 
-CMakeCache.txt
```
aardappel, line 23: why are these removed from `.gitignore` ? people may still default to in-tree builds where this is useful

aardappel (commented):
Thanks for working on this!

aardappel:
Do note that whatever you're changing has to preserve things working as before on Windows / OS X.

Not sure about Meson, I dislike CMake as much as the next guy, but it is a standard many are familiar with.

.gitignore
jtojnar: I do not see any value in in-tree builds but re-added that.

jtojnar:
> Do note that whatever you're changing has to preserve things working as before on Windows / OS X.

As I understand it, CMake is only used on Linux so it should be the same (the C++ code changes should still be compatible). Eventually, it would be nice to use CMake (or Meson) to generate VS and XCode boilerplate, and there we should follow the best practices for each individual platform, not necessarily the current behaviour.

> Not sure about Meson, I dislike CMake as much as the next guy, but it is a standard many are familiar with.

I did not realize CMake supports VS backend too, so there is a less of a pressure to switch to Meson.

aardappel:
> As I understand it, CMake is only used on Linux so it should be the same.

Yes, but like you said I'd like to keep the door open to eventually migrate over to using CMake everywhere.

CMakeLists.txt
```diff
@@ -24,11 +25,13 @@ cmake_minimum_required(VERSION 3.1)
 
 project(treesheets)
 
+include(GNUInstallDirs)
+
+set(TREESHEETS_RELOCATABLE_INSTALLATION OFF CACHE BOOL "Enable to look for data relative to the path treesheets was run from, instead of hardcoding paths from CMAKE_INSTALL_PREFIX")
```
jtojnar, line 30: I wonder if we need this option. The reason I introduced it is that I wanted to prevent relocatable builds from loading the locales from install prefix (e.g. when user has an older version installed on their system and fresh one installed into a testing directory).

src/myapp.h
```diff
@@ -21,6 +21,9 @@ struct MyApp : wxApp {
         #ifdef __WXGTK__
             locale.AddCatalogLookupPathPrefix(L"/usr");
             locale.AddCatalogLookupPathPrefix(L"/usr/local");
```
jtojnar, line 23: But the FHS paths are already hardcoded (even though they will not work due to missing `/locale` subdirectory).

CMakeLists.txt
```diff
@@ -47,6 +50,58 @@ include_directories(lobster/include lobster/src)
...
+    foreach(locale ${locales})
+        install(
+            FILES "TS/translations/${locale}/ts.mo"
+            DESTINATION "${CMAKE_INSTALL_PREFIX}/ThreeSheets.app/Contents/Resources/${locale}.lproj"
+        )
+    endforeach()
+elseif (WIN32)
+    foreach(locale ${locales})
+        install(
+            FILES "TS/translations/${locale}/ts.mo"
```
aardappel, line 89: I am not sure what this is going to do. Even if we ever had an "installer creation" on Windows, there's no need to be copying locale files, these files are tiny and the right one is automatically picked up by treesheets anyway. Better to just assume the files will get copied wholesale, and not introduce things which may break in interesting ways in the future.

aardappel: Same for os x

.github/workflows/build.yml
```diff
@@ -36,14 +36,16 @@ jobs:
         cd wxWidgets/build_
         sudo make install
     - name: cmake
-      run: CXX=${{ matrix.cxx }} cmake -DCMAKE_BUILD_TYPE=Release .
+      run: CXX=${{ matrix.cxx }} cmake -S . -B _build -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=TreeSheets-relocatable -DTREESHEETS_RELOCATABLE_INSTALLATION=ON
```
jtojnar, line 39: And also it would hardcode path like `/home/runner/work/aardappel/treesheets/TreeSheets-relocatable` into the program, which would be checked every time, even though it would not exist for users running TreeSheets.

jtojnar (commented):
I have implemented the FHS locale path and tested that the Italian translation works when installed there. There are still some questions to resolve.

.github/workflows/build.yml
aardappel: What does `relocatable` even mean here? Not familiar..

CMakeLists.txt
jtojnar: This will copy the locales to the correct path according to the platform, as described by the [wxWidgets docs](https://docs.wxwidgets.org/trunk/overview_i18n.html#overview_i18n_mofiles).

I also want to make msgfmt call part of the build as [`.mo` files are not portable](https://www.gnu.org/software/gettext/manual/html_node/Files.html). Possibly, that can cause issues on systems e.g. with different endianness.

.github/workflows/build.yml
jtojnar: Relocatable means that executing the program will work when moved to another directory. This is typically achieved by referring to resources using paths constructed at runtime by joining relative paths to `dirname(argv[0])`. Filesystem Hierarchy standard kind of implies a fixed installation prefix, which is not very relocatable.

We need to choose:

1. Support both relocatable and non-relocatable installs, look at relative path first, FHS path as a fallback.
2. Support both relocatable and non-relocatable installs, look at FHS path first, relative path as a fallback.
3. Use a configure-time switch between relocatable and non-relocatable installs. This is what the `-DTREESHEETS_RELOCATABLE_INSTALLATION` currently does.
4. Same as 1. but add configure-time switches that allow disabling each of the variants to allow Linux distros to have non-relocatable packages but still allowing PortableApps-style builds.

CMakeLists.txt
aardappel: That just shows where they would reside relative to the executable on win/osx. In both cases, they'd be in an installer or app package along with all the other standard files, and TreeSheets already knows how to access those, so there is no need for an install action on these platforms.

jtojnar: I am not very familiar with those platforms, but I would expect CMake to generate installers/bundles for those platforms (e.g. https://cmake.org/cmake/help/latest/cpack_gen/nsis.html for Windows). So if we want to use CMake on those platforms, we will also want to specify the platform-specific paths here.

aardappel: if we ever do so thru CMake, we will do the entire data dir, not just individual locale files. Please remove.

jtojnar:
Now it installs all files to their preferred directories on Linux, while (unless I missed something) keeping the old layout on Windows and Mac:

<details>
<summary>Directory tree</summary>

```nix
installed
├── bin
│  └── treesheets
└── share
   ├── applications
   │  └── treesheets.desktop
   ├── doc
   │  └── treesheets
   │     ├── docs
   │     │  ├── donations.html
   │     │  ├── file_format_spec.txt
   │     │  ├── history.txt
   │     │  ├── images
   │     │  │  ├── check.png
   │     │  │  ├── documentation
   │     │  │  │  ├── deleted.png
   │     │  │  │  ├── edit.png
   │     │  │  │  ├── grid.png
   │     │  │  │  ├── gridline.png
   │     │  │  │  ├── gridline2.png
   │     │  │  │  ├── gridsel1.png
   │     │  │  │  ├── gridsel2.png
   │     │  │  │  ├── gridsel3.png
   │     │  │  │  ├── insert.png
   │     │  │  │  ├── long1.png
   │     │  │  │  ├── long2.png
   │     │  │  │  ├── long3.png
   │     │  │  │  ├── new.png
   │     │  │  │  ├── postgrid.png
   │     │  │  │  ├── pregrid.png
   │     │  │  │  ├── textsize.png
   │     │  │  │  └── zoomed.png
   │     │  │  ├── excl.png
   │     │  │  ├── screenshots
   │     │  │  │  ├── screenshot_personel.png
   │     │  │  │  ├── screenshot_sales.png
   │     │  │  │  ├── screenshot_todo.png
   │     │  │  │  ├── screenshot_todo_half.png
   │     │  │  │  ├── screenshot_todo_linux.png
   │     │  │  │  ├── screenshot_todo_mac.png
   │     │  │  │  ├── screenshot_tutorial.png
   │     │  │  │  ├── screenshot_tutorial2.png
   │     │  │  │  └── screenshot_unicode.png
   │     │  │  ├── stop.png
   │     │  │  └── treesheets_logo.png
   │     │  ├── screenshots.html
   │     │  ├── script_reference.html
   │     │  └── tutorial.html
   │     ├── examples
   │     │  ├── complex_eval.cts
   │     │  ├── contrib
   │     │  │  ├── help
   │     │  │  │  ├── treesheet key quick reference (by WiM) 3 posted.cts
   │     │  │  │  └── treesheet menu including images landscape final.cts
   │     │  │  ├── Procrastination.cts
   │     │  │  ├── structogram-quicksort.cts
   │     │  │  ├── weekly+calendar.cts
   │     │  │  ├── what-i-like-about-treesheets.cts
   │     │  │  └── yearly-calendar.cts
   │     │  ├── imported_from_xml_examples
   │     │  │  ├── books.cts
   │     │  │  ├── cd_catalog.cts
   │     │  │  ├── menu.cts
   │     │  │  └── plant_catalog.cts
   │     │  ├── operation-reference.cts
   │     │  ├── personel_file.cts
   │     │  ├── sales.cts
   │     │  ├── simple_graph.cts
   │     │  ├── todo_calendar.cts
   │     │  ├── tutorial-de.cts
   │     │  ├── tutorial-fr.cts
   │     │  ├── tutorial.cts
   │     │  ├── unicode_test.cts
   │     │  └── work_breakdown_structure.cts
   │     └── readme.html
   ├── icons
   │  └── hicolor
   │     └── scalable
   │        └── apps
   │           └── treesheets.svg
   ├── locale
   │  ├── de
   │  │  └── LC_MESSAGES
   │  │     └── ts.mo
   │  ├── it
   │  │  └── LC_MESSAGES
   │  │     └── ts.mo
   │  └── zh_CN
   │     └── LC_MESSAGES
   │        └── ts.mo
   └── treesheets
      ├── images
      │  ├── icon16.png
      │  ├── icon32.png
      │  ├── nuvola
      │  │  ├── author
      │  │  ├── dropdown
      │  │  │  ├── apply.png
      │  │  │  ├── back.png
      │  │  │  ├── bookcase.png
      │  │  │  ├── bookmark.png
      │  │  │  ├── bug.png
      │  │  │  ├── cache.png
      │  │  │  ├── cancel.png
      │  │  │  ├── clanbomber.png
      │  │  │  ├── color_line.png
      │  │  │  ├── configure.png
      │  │  │  ├── down.png
      │  │  │  ├── edit_add.png
      │  │  │  ├── edit_remove.png
      │  │  │  ├── email.png
      │  │  │  ├── forward.png
      │  │  │  ├── gohome.png
      │  │  │  ├── history.png
      │  │  │  ├── kalarm.png
      │  │  │  ├── kcmprocessor.png
      │  │  │  ├── kcoloredit.png
      │  │  │  ├── kfm_home.png
      │  │  │  ├── kgpg.png
      │  │  │  ├── kgpg_key3.png
      │  │  │  ├── ksysv.png
      │  │  │  ├── kuser.png
      │  │  │  ├── kweather.png
      │  │  │  ├── ledblue.png
      │  │  │  ├── ledgreen.png
      │  │  │  ├── ledlightblue.png
      │  │  │  ├── ledlightgreen.png
      │  │  │  ├── ledorange.png
      │  │  │  ├── ledpurple.png
      │  │  │  ├── ledred.png
      │  │  │  ├── ledyellow.png
      │  │  │  ├── maybe.png
      │  │  │  ├── messagebox_info.png
      │  │  │  ├── messagebox_warning.png
      │  │  │  ├── misc.png
      │  │  │  ├── mozilla.png
      │  │  │  ├── no.png
      │  │  │  ├── noatunloopsong.png
      │  │  │  ├── package_games_arcade.png
      │  │  │  ├── package_network.png
      │  │  │  ├── package_toys.png
      │  │  │  ├── player_pause.png
      │  │  │  ├── player_play.png
      │  │  │  ├── player_stop.png
      │  │  │  └── up.png
      │  │  ├── fold.png
      │  │  ├── license.txt
      │  │  ├── readme.txt
      │  │  ├── thanks.to
      │  │  └── toolbar
      │  │     ├── editcopy.png
      │  │     ├── editpaste.png
      │  │     ├── filenew.png
      │  │     ├── fileopen.png
      │  │     ├── filesave.png
      │  │     ├── filesaveas.png
      │  │     ├── image.png
      │  │     ├── newgrid.png
      │  │     ├── run.png
      │  │     ├── undo.png
      │  │     ├── zoomin.png
      │  │     └── zoomout.png
      │  ├── render
      │  │  ├── line_ne.png
      │  │  ├── line_nw.png
      │  │  ├── line_se.png
      │  │  └── line_sw.png
      │  ├── treesheets.png
      │  ├── treesheets.svg
      │  └── webalys
      │     ├── readme.txt
      │     └── toolbar
      │        ├── editcopy.png
      │        ├── editpaste.png
      │        ├── filenew.png
      │        ├── fileopen.png
      │        ├── filesave.png
      │        ├── filesaveas.png
      │        ├── image.png
      │        ├── newgrid.png
      │        ├── run.png
      │        ├── undo.png
      │        ├── zoomin.png
      │        └── zoomout.png
      └── scripts
         ├── Export JSON.lobster
         └── modules
            ├── color.lobster
            ├── std.lobster
            ├── stdtype.lobster
            └── vec.lobster
```

</details>

.github/workflows/build.yml
aardappel: Well, "relocatable" is what it already does, and what it does on other platforms. I don't see why it shouldn't support that, even on Linux. So to me the only question is if it should optionally allow finding things in FHS locations as fallback.

src/myframe.h
```diff
@@ -26,9 +26,21 @@ struct MyFrame : wxFrame {
...
     ColorDropdown *borddd = nullptr;
     
-    wxString GetPath(const wxString &relpath) {
-        if (!exepath_.Length()) return relpath;
-        return exepath_ + "/" + relpath;
+    wxString GetDocPath(const wxString &relpath) {
+        #ifdef TREESHEETS_DOCDIR
+            return TREESHEETS_DOCDIR "/" + relpath;
+        #else
+            if (!exepath_.Length()) return relpath;
```
aardappel, line 33: I'd vote for always trying this first, rather than compile-time.

I'd think of the user that has an installed version, and also made a local build in their homedir to try a new feature. Likely the local build should take precedence.

CMakeLists.txt
jtojnar: Now the option just serves to toggle between the FHS-compatible layout and the layout used on Apple and Windows.

aardappel:
ok lgtm, let me know when you think its ready to merge.

src/myframe.h
jtojnar: Done.

jtojnar:
It should be good to merge now.

jtojnar:
Hmm, looks like we need to bump MACOSX_DEPLOYMENT_TARGET:

> error: 'path' is unavailable: introduced in macOS 10.15

aardappel:
Thanks for your work on this!

PR #202, opened 2022-03-12
Make the .desktop file refer to the icon by name

DeeUnderscore:
Since the svg icon is now installed into the expected location by the cmake script, the desktop file can simply refer to it by name, and it should be located accordingly.

As the file is right now, the icon will not be found, since `images/treesheets.svg` does not exist in the icon search path.


jtojnar (approved):
Oops, I must have forgotten to save this file.

aardappel:
Thanks!

PR #203, opened 2022-03-12
cmake: Split out lobster into a separate library

jtojnar:
The way the code is currently structured requires everything to be rebuilt when something small changes.

Let’s make the project more granular.

In the future, we might want to to pick up lobster externally using CMake’s find_package mechanism.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,7 +37,8 @@ set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++17")
 find_package(wxWidgets REQUIRED aui adv core xml net)
 include(${wxWidgets_USE_FILE})
 
-file(GLOB sources src/*.cpp src/*.h
+file(
+    GLOB lobster_sources
         lobster/external/flatbuffers/src/*.cpp
         lobster/src/builtins.cpp
         lobster/src/compiler.cpp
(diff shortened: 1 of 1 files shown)


aardappel:
I don't mind this change if you think it helps, but I don't understand why everything has to be rebuilt. Surely the makefile this generates doesn't rebuild the Lobster .o files if I make a change in TreeSheets proper?

As far as I understand the Linux build, this would have negligible effect on build times since all it does is resolve a few internal symbols internally to the .a, but otherwise does exactly the same work in terms compiler and linker.

Lobster tends to still have breaking API changes pretty often, so pulling in a new version of Lobster is something that has to be done "under supervision" to fix things if needed. `git submodule` may make more sense since we can pin it, just didn't want to do that since 95% of the Lobster repo is irrelevant to TreeSheets (didn't want to burden TreeSheets users with all that cruft). `git subtree` may be another option that more directly replaces my manual source file copy script. A third option is for me to refactor the Lobster repo and code better into a "language library" and "the rest" repos, but been holding off on that until there is a more pressing need.

jtojnar:
Sorry, should have been more specific. The issue I encountered was when tweaking the compiler flags (set by `target_compile_definitions` CMake function). Each time I changed them, the lobster library needed to be rebuilt, even though the defines were only relevant to the main compilation unit.

Making lobster into a library would be the best but I understand that it would be quite an undertaking.

jtojnar:
On the note of object files, currently everything is stuffed into a single main.cpp via header files, which means there is a single object file for the whole app, making compilation very slow. Would you mind if I moved the implementation from each header to a separate CPP file? (I would need someone else to handle the VS and OSX changes, though.)

aardappel:
Ok sure.. but tweaking compiler flags is not something that need optimizing :)

The actual treesheets code is really small, I very much doubt it slows compilation because it is all in headers. I'm ok with any refactoring that makes the code better, but if code is more redundant in .cpp files and doesn't compile quicker, there is no point.

jtojnar:
> Ok sure.. but tweaking compiler flags is not something that need optimizing :)

IMO, it still make sense to have a separate project like lobster as a separate target, since it is logically distinct component. And it does not even require any C++ code changes.

Tweaking compiler for the app is just what brought it to my attention since it was extremely annoying having to wait tens of seconds for lobster to recompile just to have treesheets build fail in the end, repeated many times for just minor tweaks of a single line in `CMakeScripts.txt`.



> The actual treesheets code is really small, I very much doubt it slows compilation because it is all in headers. I'm ok with any refactoring that makes the code better, but if code is more redundant in .cpp files and doesn't compile quicker, there is no point.

Building treesheets takes almost twenty seconds for me and that is a hot start with just a single line change in `myapp.h` (Less then a second spent in any other phase than `Building CXX object CMakeFiles/treesheets.dir/src/main.cpp.o`). I would expect change of this size to finish in around a second.

Unless the issue is that WxWidgets has some crazy macros, I would suspect this is because everything is in a single object file so previous builds cannot be cached.

Yeah, there will be some duplication but until we can use C++ modules, I suspect there is not much we can do. If you think that redundancy would be acceptable trade-off for fast hot start recompilation, I can try to move the implementation into cpp files and see if it is indeed the issue.

aardappel:
Ok, play around with moving code, and see what the effect is on code & compilation time.

Issue #262, opened 2022-09-14
Treesheets grabs (almost) all keyboard and mouse input in Xorg environment / gets focus even if window is unfocused


t2b3:
Seen with Treesheets on (Arch) Linux
Version (Git): r370.927f542-1
GUI Library: wxwidgets-gtk3 (3.2.1)
Desktop Environment: Xfce 4.16.1 (also reproduced with i3)

![keyboard_focus](https://user-images.githubusercontent.com/3286756/190254167-76da6fb4-37b0-43b8-b86c-a8f0f08c8004.gif)

aardappel:
I've not seen that problem in other OSes / environments, so not sure what is going there..

t2b3:
Hello! Thanks for your fast reply. I dirty-fixed this issue by adding an additional check if the window has focus

`diff --git a/src/myframe.h b/src/myframe.h
index 8facaee..0dbacf6 100755
--- a/src/myframe.h
+++ b/src/myframe.h
@@ -1046,7 +1046,7 @@ struct MyFrame : wxFrame {
     }
 
     void ReFocus() {
-        if (GetCurTab()) GetCurTab()->SetFocus();
+        if (GetCurTab() and wxWindow::HasFocus()) GetCurTab()->SetFocus();^M
     }
 
     void OnCellColor(wxCommandEvent &ce) {`

aardappel:
That sounds like a bug in your windowing system. In other places, setting the focus of a widget does that only inside the TreeSheets window, not globally.

Curious that `wxWindow::HasFocus()` does not require specifying which window...

Either way, we could merge your change if it has no adverse effects to other systems.. care to make a PR?

AntonBogun:
This resolved the issue

PR #265, opened 2022-09-17
Explicity check for window focus before refocusing

t2b3:
Under some certain circumstances in a Linux desktop environment running on a Xorg server, the Treesheets window using the WxWidgets GTK3 frontend catches all input events - even those which are not intended for Treesheets. This leads Treesheets to set the focus on its current tab and thus on its window each time an input event is detected - even if the window had no focus. Explicitly check whether the Treesheets window has current focus in the window managing system and only if it has focus, refocus on the Treesheets current tab and window.

This fixes [#262](https://github.com/aardappel/treesheets/issues/262).

### DIFF
--- a/src/myframe.h
+++ b/src/myframe.h
@@ -1046,7 +1046,9 @@ struct MyFrame : wxFrame {
     }
 
     void ReFocus() {
-        if (GetCurTab()) GetCurTab()->SetFocus();
+        if (wxWindow::HasFocus()) {
+            if (GetCurTab()) GetCurTab()->SetFocus();
+        }
     }
 
     void OnCellColor(wxCommandEvent &ce) {


aardappel:
Thanks!

aardappel:
Apparently this breaks things on Windows @AntonBogun ?

Probably easiest to do:
```
#ifdef __WXGTK__
if (wxWindow::HasFocus())
#endif
```

AntonBogun:
It does - if a treesheets has a cell selected and the view is switched with alt+tab to a different one and then back to treesheets, then it is not possible to start typing in that cell by pressing Enter and it is required to click in the window before that will be possible.
I can not build the lastest version right now to check thoroughly, but while I was on it this was the behaviour I was experiencing.

aardappel:
Ok, made this change Linux-only for now, should work as before on Windows: https://github.com/aardappel/treesheets/commit/4778c343ac78a3b3ccdaf079187b53d8cd64e235

AntonBogun:
Ok now its fixed

PR #274, opened 2022-10-20
Paint directly on Linux

### LINKED ISSUE #253: HiDPI support
Hi,

I'm running TreeSheets on a HiDPI display.

TreeSheets is usable, but not very attractive: tiny icons, text in search field too low, etc.

I can't find a good Windows' Compatibility–HiDPI setting, either.

Is that something you could improve without too much effort?

t2b3:
Instead of using a bitmap for double-buffering, draw directly like on MacOS. This avoids the case that the bitmap has to be resized for HiDPI scaling. Instead, scaling is considered during the paint.

This fixes #253 for Linux (wxGTK).


aardappel:
Thanks!

PR #275, opened 2022-10-21
Improve German translation

t2b3:
This PR includes a revision of the German translation with plenty of translated strings added. It would be very nice if you could pull this translation update.


aardappel:
Danke, gute arbeit! :)

PR #276, opened 2022-10-27
Update German translation

t2b3:
This is another iteration of the German translation. 
I would like to ask you to pull this changes. Thanks in advance!


aardappel:
Thanks!

PR #278, opened 2022-11-08
Add Brazilian Portuguese translation

georgeraraujo:
And update ancient `TS/translations/ts.pot`.


aardappel (commented):
Thanks for the updates to the main `.pot` file as well!

PR #281, opened 2022-11-09

georgeraraujo:
Fixed as described in the [Audacity wiki](https://wiki.audacityteam.org/wiki/Translating_Audacity/Writing_Translatable_Code#Dealing_with_Long_Strings).


georgeraraujo:
Here's the rationale for these changes:

I ran the command in `/src/genpot.bat` to extract the translatable strings from the source code and recompile the main template file:

https://github.com/aardappel/treesheets/blob/5997673700eff51c2338ad04a1a840a346e8a960/src/genpot.bat#L1

The generated `ts.pot` file did not detect multiline string literals. For example:

https://github.com/aardappel/treesheets/blob/5997673700eff51c2338ad04a1a840a346e8a960/src/myframe.h#L204-L205

Here's the related excerpt from the compiled template file:

https://github.com/aardappel/treesheets/blob/5997673700eff51c2338ad04a1a840a346e8a960/TS/translations/ts.pot#L617-L620

Notice that only the first line of the string was detected.

Thus, translators get a different, incomplete string to work with...

https://github.com/aardappel/treesheets/blob/5997673700eff51c2338ad04a1a840a346e8a960/TS/translations/pt_BR/ts.po#L647-L652

... which Poedit points out...

image

... and the end result is that the application does not display the translated string:

image

This pull request addresses this issue.

georgeraraujo:
Also some strings that were not marked as translatable:

https://github.com/aardappel/treesheets/blob/5997673700eff51c2338ad04a1a840a346e8a960/src/myframe.h#L274-L279

And some percent signs that are not format specifiers, which Poedit complains about:

https://github.com/aardappel/treesheets/blob/5997673700eff51c2338ad04a1a840a346e8a960/src/myframe.h#L492-L499

src/document.h
```diff
@@ -686,8 +686,8 @@ struct Document {
             int maxdepth = 0, leaves = 0;
             root->MaxDepthLeaves(0, maxdepth, leaves);
             if (maxdepth > 1)
-                return _(L"Cannot export grid that is not flat (zoom the view to the desired grid, "
-                         L"and/or use Flatten).");
+                return _(L"Cannot export grid that is not flat (zoom the view to the desired grid, \
```
aardappel, line 689: Thanks you for your comprehensive investigation to fix this. I am not sure I like this fix however, it makes source code hard to read and maintain, and no-one writes strings like this in source code. There has to be a better way. If there is no way for the tools to recognize multi-line strings of this kind correctly (why? surely recent versions of these tools must recognize modern C++?) then simply making them into single lines is a better solution for the moment, as it doesn't hurt code readability quite as much.

georgeraraujo: > Thanks you for your comprehensive investigation to fix this.

You're welcome.

I don't know whether there's a way for `xgettext` to detect multi-line strings of this kind correctly. The [documentation](https://www.gnu.org/software/gettext/manual/html_node/xgettext-Invocation.html#Output-details) states that

> This implementation of `xgettext` is able to process a few awkward cases,
> like strings in preprocessor macros, ANSI concatenation of adjacent strings,
> and escaped end of lines for continued strings.

but that did not work with the binaries for version 0.21 (the latest version is 0.21.1) I downloaded from the [site](https://mlocati.github.io/articles/gettext-iconv-windows.html) mentioned in `TS/translations/readme_translations.txt`.

I too think that there should be a better way, but couldn't find any, and while I read on the subject I found the [Audacity wiki page](https://wiki.audacityteam.org/wiki/Translating_Audacity/Writing_Translatable_Code#Dealing_with_Long_Strings) I mentioned in my [first comment](https://github.com/aardappel/treesheets/pull/281#issue-1441202645). It states that

> When strings get longer and span several lines we place a '\' at the end
> of each line, and continue immediately at the start of the next line.
> We have tried various other ways of dealing with long strings, but the
> other ways we've tried either don't work when compiling with Unicode,
> or defeat detection by xgettext. 

So I chose to do the modification that didn't use long lines.

I'll push another commit that makes those ugly multiline strings into single lines.

aardappel: That is unfortunate.. not sure how other projects deal with this..

aardappel:
Thanks for fixing!

PR #285, opened 2022-11-11
More specific loading of the localized tutorial

georgeraraujo:
First try and load the file ending with the five-letter canonical form of the current locale name, e.g. `tutorial-pt_BR.cts`. Failing that, try and load the file ending with the two-letter canonical form of the current locale name, e.g. `tutorial-pt.cts`. Failing that, just load `tutorial.cts`, and be done with it.

### DIFF
--- a/src/system.h
+++ b/src/system.h
@@ -209,13 +209,16 @@ struct System {
     void LoadTut() {
         auto lang = frame->app->locale.GetCanonicalName();
 
-        if (lang.Len() == 5 && LoadDB(frame->GetDocPath(L"examples/tutorial-" + lang + ".cts"))[0]) {
-            lang.Truncate(2);
+        if (lang.Len() == 5 && !LoadDB(frame->GetDocPath(L"examples/tutorial-" + lang + ".cts"))[0]) {
+            return;
         }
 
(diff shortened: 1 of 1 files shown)


src/system.h
```diff
@@ -208,8 +208,12 @@ struct System {
...
     void LoadTut() {
         auto lang = frame->app->locale.GetCanonicalName();
-        lang.Truncate(2);
-        if (LoadDB(frame->GetDocPath(L"examples/tutorial-" + lang + ".cts"))[0]) {
+
+        if (lang.Len() == 5 && LoadDB(frame->GetDocPath(L"examples/tutorial-" + lang + ".cts"))[0]) {
+            lang.Truncate(2);
+        }
+
+        if (lang.Len() == 2 && LoadDB(frame->GetDocPath(L"examples/tutorial-" + lang + ".cts"))[0]) {
```
aardappel, line 216: I guess the difference is the old code would work even if `lang` was empty etc. Recommend making this code more robust by doing something along these lines:
```
if (len == 5 && !LoadDB..) return;  // Success with 5
truncate(2);
if (len == 2 && !LoadDB..) return;  // Success with 2
LoadDB..  // Load default.
```

georgeraraujo: Done.

aardappel:
Thanks!

PR #287, opened 2022-11-13
 pt_BR translation: fix UI keyboard shortcut...

georgeraraujo:
... and update tutorial.
Also recompile the main template file, and update the PO file for all other languages (but not regenerate the MO file, that is up for translators to do.)


aardappel:
The updated PO for other languages also adds some translations, and at least for DE I can see some seem to be incorrect, e.g.

```
msgid "Border &0\tCTRL+SHIFT+9"
msgstr "Alles einklappen\tCTRL+SHIFT+F10"
```

Not only does the the translation not match (it says "fold everything"), it also has the wrong keyboard shortcut.

aardappel:
@t2b3 who recently worked on the DE translation.

georgeraraujo:
> I am guessing it is taking the translation from an existing item?

Yes, it is a "fuzzy" translation in gettext parlance. Poedit highlights those:

image

The MO file (the one that contais the actual translated strings) remains untouched.

aardappel:
I don't think we want these "fuzzy" translations in there though, if they're wrong. If they get recompiled without the language author noticing them, we have bad strings in the UI.

georgeraraujo:
> I don't think we want these "fuzzy" translations in there though, if they're wrong. If they get recompiled without the language author noticing them, we have bad strings in the UI.

While the the language author might run `TS/translations/<language>/merge.bat` and not notice those fuzzy strings, they might just as well not notice they're translating a `TS/translations/<language>/ts.mo` file generated from an outdated `TS/translations/ts.pot`. That's precisely what happened with the zh_CN translation, where in November '19 the author translated strings dating back to March '17 (`TS/translations/<language>/merge.bat` simply did not exist back then, I only pushed it a few days ago):

https://github.com/aardappel/treesheets/blob/94881402b4730f598a536d8cddb3314eaf16defc/TS/translations/zh_CN/ts.po#L10-L11

My intention was that upon seeing the highlighted fuzzy strings in their translation tool of choice, the language author would translate them.

aardappel:
Some people might not use a translation tool that highlights these problems, but simply use a text editor and not notice them?

georgeraraujo:
If they ignore the `#, fuzzy` line msgmerge inserts to flag an entry as fuzzy , then yes.

image

aardappel:
I don't know, I've never worked on translations myself, but I can imagine that would be a problem.

PR #289, opened 2022-11-14
Update German translation to match updated POT

t2b3:
Sync with the latest updates of the Translation Template file from @georgeraraujo


aardappel:
Thanks!

PR #290, opened 2022-11-22
Init translate french

Philippe-devel-github:


aardappel:
Merci!

georgeraraujo:
Great work @Philippe-devel-github !

Just so that you don't miss any new strings, remember to run the command in `treesheets/src/genpot.bat` in order to recompile the main template file (extract the translatable strings from the source code), and then merge it into your translation with `treesheets/TS/translations/fr_FR/merge.sh`. That way you'll have the most current translatable strings to work with.

PR #297, opened 2022-12-07
Provide original PNG as wxCustomDataObject for the clipboard

t2b3:
When available, use the original PNG data instead of the bitmap for the clipboard.


aardappel:
nice!

PR #299, opened 2022-12-08
Provide original PNG to clipboard only on Linux

t2b3:
For other platforms, just provide the bitmap object. This should fix #297 on Windows.

### DIFF
--- a/src/document.h
+++ b/src/document.h
@@ -1553,13 +1553,16 @@ struct Document {
                 if (selected.Thin()) return NoThin();
 
                 if (wxTheClipboard->Open()) {
+                    #ifdef __WXGTK__
                     if (!c->text.image->png_data.empty()) {
                         wxCustomDataObject* pngimage = new wxCustomDataObject(wxDF_BITMAP);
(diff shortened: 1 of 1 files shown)


aardappel:
Checked on Windows, and it crashed because it was the wrong format. Not sure why Linux doesn't seem to care? :)
https://github.com/aardappel/treesheets/commit/39e51b894a7b64bb649a42717702484a69ddf934

t2b3:
Thanks for fixing this! :-)

aardappel:
Hmm it says "only valid on MSW".. does it still work on Linux?

t2b3:
Unfortunately not, it seems to fail:
image

t2b3:
For the time being, I created https://github.com/aardappel/treesheets/pull/300 as fix that distinguishes between Microsoft Windows platform and the others (as wxDF_PNG is not available on Linux or Mac).

aardappel:
Wow what a mess, thanks wxWidgets: https://github.com/aardappel/treesheets/commit/755494e0e12ecbc8628bbb7f75acd2d72e641b3e

PR #300, opened 2022-12-08
Treat original PNG as wxDF_PNG or wxDF_BITMAP

t2b3:
According to
<https://docs.wxwidgets.org/latest/classwx_data_format.html>, wxDF_PNG is only available for MSW, but not for Linux or Mac. Use wxDF_BITMAP for non-Windows platforms.


aardappel:
Sorry I made a fix at the same time. Also left out `__WXMAC__` since its untested if it supports this path.

PR #301, opened 2022-12-10
Some more convenience functionalities (Copy Image, allow wxDF_PNG also for pasting, Paste As Continuous Text)

t2b3:
Dear Wouter

I thought that maybe we can add a little more convenience for the users with regards to the copy functionality.

If only one cell is selected and this cell has an image and no text is in the cell, then copy the image to clipboard.
If otherwise, copy text to clipboard.

Regards,
Tobias

### DIFF
--- a/src/document.h
+++ b/src/document.h
@@ -1331,7 +1331,9 @@ struct Document {
                 }
                 return nullptr;
 
+            case A_PASTECT:
             case A_PASTE:
+                k == A_PASTECT ? pastect = true : pastect = false;
                 if (!(c = selected.ThinExpand(this))) return OneCell();
                 if (wxTheClipboard->Open()) {
                     wxTheClipboard->GetData(*dataobjc);
@@ -1341,6 +1343,7 @@ struct Document {
                     c->Paste(this, sys->cellclipboard.get(), selected);
                     Refresh();
                 }
+                pastect = false;
                 return nullptr;
 
             case A_PASTESTYLE:
@@ -1739,6 +1742,8 @@ struct Document {
 
     void PasteSingleText(Cell *c, const wxString &t) { c->text.Insert(this, t, selected); }
 
+    bool pastect = false;
+
     void PasteOrDrop() {
         Cell *c = selected.ThinExpand(this);
         if (!c) return;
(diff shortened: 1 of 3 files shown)


t2b3:
I had to add some small check that we ensure that only one cell is selected if we want to copy an image.

t2b3:
Also allow wxDF_PNG as image format for pasting from clipboard as PNG images are also copied to the clipboard.

t2b3:
Also add functionality to paste string in clipboard as continous text (no split).

src/document.h
```diff
@@ -830,6 +830,29 @@ struct Document {
...
+                );
+                pngimage->SetData(cell->text.image->png_data.size(), cell->text.image->png_data.data());
+                wxTheClipboard->SetData(pngimage);
+            } else
+            #endif
+            {
+                wxTheClipboard->SetData(new wxBitmapDataObject(cell->text.image->bm_orig));    
+            }
+            wxTheClipboard->Close();
+            }
```
aardappel, line 852: wrong indent here

src/document.h
```diff
@@ -1302,7 +1331,9 @@ struct Document {
                 }
                 return nullptr;
 
+            case A_PASTECT:
```
aardappel, line 1334: why is this a new separate operation?

src/document.h
```diff
@@ -1302,7 +1331,9 @@ struct Document {
                 }
                 return nullptr;
 
+            case A_PASTECT:
             case A_PASTE:
+                k == A_PASTECT ? pastect = true : pastect = false;
```
aardappel, line 1336: assignment outside conditional

src/document.h
```diff
@@ -1312,6 +1343,7 @@ struct Document {
                     c->Paste(this, sys->cellclipboard.get(), selected);
                     Refresh();
                 }
+                pastect = false;
```
aardappel, line 1346: why is this not just an argument to `PasteOrDrop` ?

src/document.h
t2b3: Thanks for taking your time! I corrected this in https://github.com/aardappel/treesheets/pull/301/commits/c84f3d994a415871736a6918c569f6e943116d66

src/document.h
t2b3: Thanks for this hint! I made it an optional argument to PasteOrDrop.

src/document.h
aardappel: Yes, but why do we have this new operation at all? If not frequently needed it just clutters the menu/code, and there's other ways to work around it, like use the existing contiguous copy.

t2b3: Thank you for your reply. I use it a lot when I want to copy originally multi-line textes to one cell. Otherwise I have to copy them line by line to the clipboard and paste them from the clipboard. This would save some time.

PR #303, opened 2022-12-10
Copy image if only image is in cell, small fixes

t2b3:
Hello Wouter

This is an iterative approach. I stripped the "Paste As Continous Text" functionality away.
This pull requests only encompasses the possibility to copy the image directly if only the image is in the cell.
Also it added some small fix for also supporting wxDF_PNG for pasting from the clipboard and fixes the indentation issue.

Kind regards,
Tobias

### DIFF
--- a/src/document.h
+++ b/src/document.h
@@ -830,6 +830,29 @@ struct Document {
         return nullptr;
     }
 
+    const wxChar* CopyImageToClipboard(Cell *cell) {
+        if (wxTheClipboard->Open()) {
+            #if defined(__WXMSW__) || defined(__WXGTK__)
+            if (!cell->text.image->png_data.empty()) {
+                wxCustomDataObject *pngimage = new wxCustomDataObject(
+                    #ifdef __WXMSW__
+                        wxDF_PNG
+                    #else
+                        wxDF_BITMAP
+                    #endif
+                );
+                pngimage->SetData(cell->text.image->png_data.size(), cell->text.image->png_data.data());
+                wxTheClipboard->SetData(pngimage);
+            } else
+            #endif
+            {
+                wxTheClipboard->SetData(new wxBitmapDataObject(cell->text.image->bm_orig));    
+            }
+            wxTheClipboard->Close();
+            }
+        return _(L"Image copied to clipboard");
+    }
+
     const wxChar *Action(wxDC &dc, int k) {
         ShiftToCenter(dc);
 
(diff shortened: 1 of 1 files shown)


src/document.h
```diff
@@ -830,6 +830,23 @@ struct Document {
         return nullptr;
     }
 
+    const wxChar* CopyImageToClipboard(Cell *cell) {
+        if (wxTheClipboard->Open()) {
+            #ifdef __WXGTK__
```
aardappel, line 835: why does it not work on Windows? seemed to work for me..

t2b3:
I tested under Windows 10 with Microsoft Office Word and no image was
pasted from the clipboard although the original PNG image was (said to have
been) copied to clipboard from TreeSheets. With the wxBitmap from bm_orig
it is working on Windows.

Wouter van Oortmerssen ***@***.***> schrieb am So., 11. Dez.
2022, 00:58:

aardappel:
I don't have Word, but I tested with the current code, and was able to copy image (and verified it goes thru the PNG specific code), and then paste that image into Chrome (gmail) for example.

So it works generally, but seems Word doesn't like PNG data in the clipboard whereas Chrome doesn't mind.

For the future, please make PRs addressing a single concern, that makes them quicker to review and merge. You had a PR that mixed copying images with line merging, and now you have a PR that mixes copying images with changing windows behavior of it. The former functionality could have been merged long ago if you don't keep adding things which need a possibly separate review. It also means that if in the future we discover it is buggy it is easy to undo.

aardappel:
Ok, tried a few other apps, and so far Chrome is the only one that manages to read the data? So let's merge this for now.

Would have been nice to at leave some comments in the code about the Window situation (that it only works in some apps, and that if turned on, it would need the PNG constant.

t2b3:
> For the future, please make PRs addressing a single concern, that makes them quicker to review and merge. You had a PR that mixed copying images with line merging, and now you have a PR that mixes copying images with changing windows behavior of it. The former functionality could have been merged long ago if you don't keep adding things which need a possibly separate review. It also means that if in the future we discover it is buggy it is easy to undo.

Thank you for your time and your review! I understand this and I will take your remarks into consideration for the future.

t2b3:
> Ok, tried a few other apps, and so far Chrome is the only one that manages to read the data? So let's merge this for now.
> 
> Would have been nice to at leave some comments in the code about the Window situation (that it only works in some apps, and that if turned on, it would need the PNG constant.

The situation seems to be bit confusing. Following your remarks, I should let this PR untouched for merging?

aardappel:
Yes, sorry, should have merged it :)
Thanks!

PR #305, opened 2022-12-11
Update FSF address

jn64:
I only touched user-facing files. Nuvola project is defunct so it can only be updated here.

Other files in this repo with the old address are from libtcc via lobster.

[LGPL 2.1 on gnu.org with current address](https://www.gnu.org/licenses/old-licenses/lgpl-2.1-standalone.html)


aardappel:
I don't think this matters, but I don't mind merging it.

jn64:
I agree, but I would like to eventually package this in Fedora and this is picked up by the linter :)

https://duckduckgo.com/?q=incorrect-fsf-address

PR #307, opened 2022-12-12
Reset layout when resetting image scale

t2b3:
When the image scale is reset, also reset the layout to adapt the cell to the new dimensions of the image.


aardappel:
Thanks :)

PR #311, opened 2022-12-14
Francais

Philippe-devel-github:
After rebase
    Add a second "color cell" in the toolbar
to change faster the color of cells with two color selected

    may use a parameter "csf_lu" in the config file to choose a content scale different of system
with the high resolution i think you may need to have a csf different in the application. with this modification you can put a parameter csf_lu to change it

    icons in the toolbar adapt to the csf
if you have csf different csf_orig i think better with csf


aardappel:
Like I said before, please separate unrelated changes into individual PRs. One PR per feature/change.

Philippe-devel-github:
ok

aardappel:
https://github.com/aardappel/treesheets/issues/328

PR #312, opened 2022-12-15
csf icon size

Philippe-devel-github:
icons in the toolbar adapt to the csf

if you have csf different csf_orig i think better with csf

### DIFF
--- a/src/myframe.h
+++ b/src/myframe.h
@@ -659,7 +659,7 @@ struct MyFrame : wxFrame {
             auto AddTBIcon = [&](const wxChar *name, int action, wxString file) {
                 wxBitmap bm;
                 if (bm.LoadFile(file, wxBITMAP_TYPE_PNG)) {
-                    auto ns = csf_orig * sc;
+                    auto ns = csf * sc;
                     ScaleBitmap(bm, ns, bm);
                     MakeInternallyScaled(bm, tb->GetBackgroundColour(), csf_orig);
                     tb->AddTool(action, name, bm, bm, wxITEM_NORMAL, name);


Philippe-devel-github:
Good morning
on my laptop linux High res
csf_orig = 2
csf = 1 
version before
image

version after 
image

aardappel:
How does this work on Linux? According to the code, `csf` and `csf_orig` are always the same on that platform: 
https://github.com/aardappel/treesheets/pull/312/files#diff-703dcda1b1797e283f9c36492f6c58a9048dcf1fcec1298842021585e2c09f10R124-R146
Also, not sure if the 2nd screenshot is better, everything is now too small, including the other ui.

Philippe-devel-github:
No not the same 
it's the same on old desktop but if you have DPI devices you must change csf if you want to see something specially on laptop.
On my laptop i have csf = 2,then icons are too big in my opinion if i compare with another application
i have another pull request on the future to work with a csf on param file to have a scale you want on treesheets

aardappel:
Where are these two variables set to different values in the current code?

If this PR depends on a future PR, that PR needs to be merged first, because currently this change does nothing?

Philippe-devel-github:
this PR not depend of the other.... Icons are too big for a High DPI on laptop

Philippe-devel-github:
you can see also that in the first screenshot that the toolbar is incomplete because of the icon size and complete in the second
You say that the second screenshot is too small but it isn't true with high dpi screen with csf = 2

aardappel:
Can you point me where in the code `csf` and `csf_orig` get different values?

Philippe-devel-github:
i agree with you ..... in your code it's the same 
It's difficult to explain for me i'm not "english spoken fluently"
i try another time
because of High DPI, my content scale in the system (GetContentScaleFactor) is 2
so csf = 2
and with csf = 2, if you look at the first screenshot without modification i haven't "border" and "Image" controls in screen but out of screen
Not very well

To have all in the screen - in my fork (and only in my fork) 
at this part 
`        #ifdef __WXGTK__
            // Currently on Linux csf is always 1, so this is useless.
            // FIXME: On a high-DPI display we get low res images even though the display is
            // capable of better!
        #endif
`
i add a line 
'csf=1'

(I don't propose this in PR because i'm not sure that is the same thing for everyone)

and this is the result  :
image

you can see one more control "border" but not all, the image control is out of screen

You are not agree but i say that in this configuration csf_orig = 2, Icons are too big
So i change with csf 
and you can see in the screenshot "after" that i have all controls with icons not too big

in your code the csf and csf_orig are the same .... for the moment
if you (or anyone) change something in your code 
i think in particular in this zone in the code :
`        #ifdef __WXMSW__
            // On Windows, I get csf == 1.25, as indicated in the display properties.
            // With this factor set, bitmaps display. At their same physical sizes as when
            // TreeSheets was a non-DPI-aware app, and extra resolution is used.
        #endif
        #ifdef __WXMAC__
            // Typically csf == 2 on a retina mac. But on the mac, unlike Windows, image rendering
            // *already* does scaling, and no way to turn that behavior off for now?
            csf = 1.0;   
            // FIXME: This gives us low res images even though the display is capable of better!
            // Apparently still not fixed: http://trac.wxwidgets.org/ticket/15808
            // wxBitmap::CreateScaled could be the way to solve this, but it is not obvious
            // how to use it, since you can't pass this scale to LoadFile etc. Could possibly
            // blit it over via a MemoryDC?
        #endif
        #ifdef __WXGTK__
            // Currently on Linux csf is always 1, so this is useless.
            // FIXME: On a high-DPI display we get low res images even though the display is
            // capable of better!
        #endif
`
the csf can change and be different of csf_orig and 
i don't know how to say it i think it is more in the spirit to put csf than csf_orig

If you aren't agree tell me and i close the PR

PR #314, opened 2022-12-16
Add "Go to next filter match"

t2b3:
Use Case: Sometimes cells with filter matches are highly dispersed throughout the document. Make it easier to find them by providing this functionality.

Also add an additional check on the parent cell (see document.h).

This avoids an invalid pointer causing a segfault when jumping between cells that are filtered by their last edits.


src/document.h
```diff
@@ -1716,6 +1719,19 @@ struct Document {
         return nullptr;
     }
 
+    const wxChar *FilterMatchNext(wxDC &dc) {
```
aardappel, line 1722: Please inline this function

t2b3: Thanks for your review! I revised the commit.

aardappel:
Thanks!

PR #320, opened 2022-12-20
Make filter off more generic

t2b3:
If one can filter by search term or by cell color, turning the filter off should be more generic.

### DIFF
--- a/src/document.h
+++ b/src/document.h
@@ -1107,7 +1107,11 @@ struct Document {
                 ApplyEditFilter();
                 return nullptr;
             case A_FILTERS: SetSearchFilter(true); return nullptr;
-            case A_FILTEROFF: SetSearchFilter(false); return nullptr;
+            case A_FILTEROFF: 
+                loopallcells(c) c->text.filtered = false;
+                rootgrid->ResetChildren();
+                Refresh();
+                return nullptr;
 
             case A_CUSTKEY: {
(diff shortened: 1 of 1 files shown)


src/document.h
```diff
@@ -1107,7 +1107,11 @@ struct Document {
                 ApplyEditFilter();
                 return nullptr;
             case A_FILTERS: SetSearchFilter(true); return nullptr;
-            case A_FILTEROFF: SetSearchFilter(false); return nullptr;
+            case A_FILTEROFF: 
+                loopallcells(c) c->text.filtered = false;
```
aardappel, line 1111: This appears to just copy code from the body of that function without any change in functionality? Can you tell me what this is supposed to do? And if you want to change something, better to do so inside the function.

t2b3: This is right. I just thought that is should look more generic because the function suggests that it just deals with the search filter not other filters (e.g. by cell color). But the function remains the same. So I will close this PR for now.

aardappel: sorry, still don't understand the intention.. it makes the code a lot less maintainable by having 2 copies of the same.. its less generic in an abstraction sense..

t2b3:
This PR just duplicates code. So I close it.

src/document.h
t2b3: Better I had had second thoughts about it...

PR #321, opened 2022-12-20
Fix memory leak when another instance is running

t2b3:
The instance_checker as a wxSingleInstanceChecker class instance shall be destroyed before exiting the program. This is in compliance with the class reference.

Memory leak found with the AddressSanitizer bundled with Clang.


aardappel:
Thanks!

PR #322, opened 2022-12-30
Apply open link / open file to multiple files

t2b3:
Loop through all selected cells and open the files / links

### DIFF
--- a/src/document.h
+++ b/src/document.h
@@ -1536,6 +1536,43 @@ struct Document {
                 Refresh();
                 return nullptr;
             }
+
+            case A_BROWSE: {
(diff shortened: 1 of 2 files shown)


aardappel:
hmm.. potentially dangerous if you have a large selection and its going to open 100s of links and freeze your system.. I'd say it be good to bail out after N link opens.. not sure what N should be.. possibly 10 or 20 or so?

t2b3:
Thank you for your review! I added a counter for application launches and if it exceeds, the program breaks the loop. What do you think about it?

aardappel:
Nice, thanks.

PR #329, opened 2023-01-04
Add filter and search icons to toolbar

t2b3:
- Adds search icon next to search field
- Adds filter icon next to search field
- Adds filter icon next to cell background
- Adds go-next icon for filter
- Adds filter-off icon for filter


aardappel:
Some of these icons look very low-res.. please make sure to find appropriate ones from the indicated icon sets (nuvola etc).

Can you post a screenshot of what the new tool bar looks like?

t2b3:
Thanks for your reply! I will search for the appropriate sizes. I searched in the icon themes but could not find the right sizes. I will further investigate.

It would look like this:
image

t2b3:
I could not find 48x48 filter icons from the nuvola icon set. In kdeartwork (15.08) there is the mono and the primary icon themes (SVG needed to be rasterized). Are they a worthy replacement?

aardappel:
While the search button addition is probably welcome, we have to be quite careful about the size of the toolbar, since a lot of people on small screens in the past have complained of buttons not being reachable.

I am not sure that I would would want any of the filter actions in the toolbar.. this is an extremely niche experimental feature that I am not sure many people are actually using.

As for the icons, being higher res is certainly most important.. it be good to pick something from nuvola for style consistency.

t2b3:
> While the search button addition is probably welcome, we have to be quite careful about the size of the toolbar, since a lot of people on small screens in the past have complained of buttons not being reachable.
> 
> I am not sure that I would would want any of the filter actions in the toolbar.. this is an extremely niche experimental feature that I am not sure many people are actually using.
> 
> As for the icons, being higher res is certainly most important.. it be good to pick something from nuvola for style consistency.

Thanks for your reply! I will strip the filter icons away and create a new PR.

PR #333, opened 2023-01-22
[DRAFT/NO MERGE] Consider Wayland scaling

t2b3:


src/myframe.h
```diff
@@ -140,9 +140,17 @@ struct MyFrame : wxFrame {
...
+            // On X11 platform, there is no scaling factor beside 1.0, but the DPI can be changed.
+
+            // On Wayland platform, there might be a scaling factor beside 1.0, 
+            // but the DPI is assumed to be a fixed value by default.
+
+            // If upscaling is active in Wayland compositors, the reported bitmap to paint on is downsized (and vice versa).
+            // The scaling is handled by the GTK toolkit and the Wayland compositor. 
+            // This is why csf and csf_orig should be set to 1.0 to let scaling be handled by the
+            // toolkit and compositor.
+            csf_orig = 1.0;
```
aardappel, line 152: Why change `csf_orig` ? the whole point of that variable is that it is the original value returned by `GetContentScaleFactor`.

So on Wayland, does `GetContentScaleFactor` return other values than 1.0?

t2b3: Thanks for looking into this! I removed `csf_orig`. Yes, under Wayland compositors and scaling active, the active scaling factor is reported.

aardappel:
Ok, thanks!

t2b3:
> Oh, it still says "no merge" in the title.. did you want to change anything about this?
> 
> I'm not familiar with Linux scaling mechanisms so your guess is as good as any, unless anyone else on Linux wants to chime in what the best strategy is.

Well I think it is good to go.

PR #334, opened 2023-01-23
Add mime info

### LINKED ISSUE #178: .cts files in Linux do not seem to have an assigned Magic Number
**In Linux, installations need to have the Magic Number to .cts files added.**

I believe the file starts with **"TSFF\n"**, but please do correct me if I am wrong. Without it, .cts files are listed as _"unknown type"_ and it is **NOT possible to open a .cts file by double clicking on the link**, or alternatively **ALL** files of the type _"unknown"_ are being open by Treesheets (which is not the _desirable_ outcome).

I am not sure who is empowered to request it.

jn64:
Closes #178

- On Linux, the mime xml file is typically installed to `/usr/share/mime/packages/treesheets.xml`.

- Users who manually install/build TreeSheets may need to update the mime database:

  ```sh
  # update-mime-database /usr/share/mime
  ```

  Downstream packages should take care of this during installation.

- *.cts files should be automatically associated with TreeSheets program [via the desktop file](https://github.com/aardappel/treesheets/blob/4d48bbe730b19963efeb3bb6f59e92aa3968b229/TS/treesheets.desktop#L12), users do not need to "Open with" TreeSheets manually like in [this comment](https://github.com/aardappel/treesheets/issues/178#issuecomment-1019578780).

- Translators can translate "TreeSheets document" in the xml file. [See here for an example](https://www.freedesktop.org/wiki/Specifications/AddingMIMETutor/).

---

Screenshot of my file manager after applying this:

screenshot_2023-01-23_11-40-28


aardappel:
Thanks :)

PR #336, opened 2023-02-06
Use pass-by-reference for cell looping and image resize functions

t2b3:
This avoids the copy of variables and pointers in the case of cell looping in general and in addition when calling the methods for images resize.

### DIFF
--- a/src/document.h
+++ b/src/document.h
@@ -89,13 +89,13 @@ struct Document {
 
     #define loopcellsin(par, c) \
         CollectCells(par);      \
-        loopv(_i, itercells) for (Cell *c = itercells[_i]; c; c = nullptr)
+        loopv(_i, itercells) for (Cell *&c = itercells[_i]; c; c = nullptr)
     #define loopallcells(c)     \
         CollectCells(rootgrid); \
-        loopv(_i, itercells) for (Cell *c = itercells[_i]; c; c = nullptr)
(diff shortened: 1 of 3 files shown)


src/system.h
```diff
@@ -19,18 +19,18 @@ struct Image {
     Image(wxBitmap _bm, uint64_t _hash, double _sc, vector<uint8_t> &&pd)
         : bm_orig(_bm), png_data(std::move(pd)), hash(_hash), display_scale(_sc) {}
 
-    void BitmapScale(double sc) {
+    void BitmapScale(const double &sc) {
```
aardappel, line 22: This is not an improvement.. both `double` and a `&` are 64-bit arguments and the latter is slower to access.

src/document.h
```diff
@@ -89,13 +89,13 @@ struct Document {
 
     #define loopcellsin(par, c) \
         CollectCells(par);      \
-        loopv(_i, itercells) for (Cell *c = itercells[_i]; c; c = nullptr)
+        loopv(_i, itercells) for (Cell *&c = itercells[_i]; c; c = nullptr)
```
aardappel, line 92: What is this for? The only thing this would do is allow the loop code to modify `c`, but we don't have any use for that currently.

t2b3:
Thank you for your time and review! I will take the PR back.

Issue #200, opened 2022-02-11
Accessible display of search highlights


despens:
At the moment, TreeSheets highlights text on the sheet that matches input into the search box by using red type on white background.

People with a color vision deficiency ([deuteranopia and deuteranomaly](https://en.wikipedia.org/wiki/Color_blindness#Deuteranopia)) have a very hard time or won't be able to distinguish this highlighting from the default black text on white background.

I cannot see the red highlight here:
![image](https://user-images.githubusercontent.com/571494/153559254-1c8f387a-74e4-4f4d-81ae-2df2084c0952.png)

My suggestion would be instead of highlighting the matches, reducing contrast for cells that do not match (mockup):
![image](https://user-images.githubusercontent.com/571494/153559837-9515d2e6-bc84-4dde-8f5f-5227fd7c6643.png)

This "torchlight" is a very accessible way of highlighting results without reordering a list, as used for instance in Mac OS system preferences:
![image](https://user-images.githubusercontent.com/571494/153560396-3e8de7dd-9c69-4591-a89a-b3683164fa73.png)

aardappel:
I agree that be a cool feature!

despens:
Wow this is amazing! Thanks @t2b3 and @aardappel :pray:

PR #341, opened 2023-02-22
Darken non-matching cells in search

t2b3:
This PR adds the possibility to darken non-matching cells in search


src/cell.h
```diff
@@ -141,6 +141,12 @@ struct Cell {
                 p = p == doc->curdrawroot ? nullptr : p->parent;
             if (p) parentcolor = p->actualcellcolor;
         }
+
+        if(sys->_darkennonmatchingcells && !this->text.IsInSearch()) {
+            uchar *cp = (uchar *)&actualcellcolor;
+            loop(i, 4) cp[i] = cp[i] * 800 / 1000;
```
aardappel, line 147: This is very slow doing this on a per cell basis. First, this is another check of it being in search, which needs to do a string search. Second, this potentially changes the color of all cells, requiring a lot of extra drawing.

This seems a very niche feature, and I would not agree it should be an option. It's either always on how we display search, or not. I vote for not since its expensive.

Generally, when adding features, consider there is a cost to adding "yet another option" to the UI. The UI should be kept as simple as possible, and only options for when its a thing likely a lot of people will want to interact with.

t2b3: Thanks for your feedback! I stumbled across https://github.com/aardappel/treesheets/issues/200 and thought it would be a nice feature as you also agreed upon in the issue. Should it not be implemented?

aardappel: That is not the only issue, see my comments above.
We can maybe implement it if it can be made efficient (certainly when no search is happening) and likely just always on.

t2b3: Ok, I might come up with a second iteration.

PR #342, opened 2023-02-23
Jump to next match when pressing enter in search

t2b3:
Use case:
* Press Ctrl+F to activate search field
* Type in search term
* Press enter key to jump to first match
* Press enter key again to jump to next match
* ... and so forth ...


aardappel:
Nice!

PR #343, opened 2023-02-23
Add Save Image(s) to disk functionality

t2b3:
Use-case:
* Saving images directly from TreeSheets to a location the user wants

If multiple images are in the selection, the file name of the images after the first one will be appended by a counter.


aardappel:
Thanks!

PR #344, opened 2023-02-26
Remember last selected entity

t2b3:
When the TreeSheets file is reopened, the selected cell is the one selected when the TreeSheets file was saved.

Use case: In a TreeSheets file, the user can continue right were the user stopped working before.

### DIFF
--- a/src/cell.h
+++ b/src/cell.h
@@ -300,20 +300,24 @@ struct Cell {
         doc->AddUndo(this);
     }
 
-    void Save(wxDataOutputStream &dos) const {
+    void Save(wxDataOutputStream &dos, Cell *ocs) const {
         dos.Write8(celltype);
         dos.Write32(cellcolor);
         dos.Write32(textcolor);
         dos.Write8(drawstyle);
+        uint cellflags = (this == ocs) ? TS_SELECTION_MASK : 0;
         if (HasTextState()) {
-            dos.Write8(grid ? TS_BOTH : TS_TEXT);
+            cellflags |= grid ? TS_BOTH : TS_TEXT;
+            dos.Write8(cellflags);
             text.Save(dos);
-            if (grid) grid->Save(dos);
+            if (grid) grid->Save(dos, ocs);
         } else if (grid) {
-            dos.Write8(TS_GRID);
-            grid->Save(dos);
+            cellflags |= TS_GRID;
+            dos.Write8(cellflags);
+            grid->Save(dos, ocs);
         } else {
-            dos.Write8(TS_NEITHER);
+            cellflags |= TS_NEITHER;
+            dos.Write8(cellflags);
         }
     }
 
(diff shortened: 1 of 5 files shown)


src/main.cpp
```diff
@@ -32,7 +32,7 @@ static uint celltextcolors[] = {
 };
 #define CUSTOMCOLORIDX 0
 
-enum { TS_VERSION = 20, TS_TEXT = 0, TS_GRID, TS_BOTH, TS_NEITHER };
+enum { TS_VERSION = 21, TS_TEXT = 0, TS_GRID, TS_BOTH, TS_NEITHER, TS_TEXT_LS, TS_GRID_LS, TS_BOTH_LS, TS_NEITHER_LS };
```
aardappel, line 35: Rather than all these variants, which create a lot of complexity in code, just define the upper bit of the byte as the selection:

TS_SELECTION_MASK = 0x80

if (byte & TS_SELECTION_MASK) // here we know the cell is selected.

byte &= ~TS_SELECTION_MASK  // clear the bit, so that the rest of the code can remain unchanged.

t2b3: Thanks for your taking your time and for your feedback.
Would the revised commit correspond to your idea?

src/cell.h
```diff
@@ -300,20 +300,20 @@ struct Cell {
...
 
-    void Save(wxDataOutputStream &dos) const {
+    void Save(wxDataOutputStream &dos, Cell *ocs) const {
         dos.Write8(celltype);
         dos.Write32(cellcolor);
         dos.Write32(textcolor);
         dos.Write8(drawstyle);
         if (HasTextState()) {
-            dos.Write8(grid ? TS_BOTH : TS_TEXT);
+            dos.Write8(grid ? ((this == ocs) ? TS_BOTH | TS_SELECTION_MASK : TS_BOTH) : ((this == ocs) ? TS_TEXT | TS_SELECTION_MASK : TS_TEXT));
```
aardappel, line 309: why do this twice? just create the flags first in a var, then `|` if selected.

src/document.h
```diff
@@ -137,9 +137,20 @@ struct Document {
...
-    void InitWith(Cell *r, wxString filename) {
+    void InitWith(Cell *r, wxString filename, Cell *ics) {
         rootgrid = r;
-        selected = Selection(r->grid, 0, 0, 1, 1);
+        if(ics) {
+            Grid* ipg = ics->parent->grid;
+            if(ipg) {
+                foreachcellingrid(c, ipg) {
+                    if(c == ics) {
+                        selected = Selection(ipg, x, y, 1, 1);
```
aardappel, line 147: it's going to be slightly odd this only works with 1x1 selections? maybe you can save the selection size in the file (be sure to check with the version on read)

src/cell.h
t2b3: I revised the commit. Is this ok now?

t2b3: I revised this. Is this ok now?

src/document.h
aardappel: It is very little extra code. And I think it makes for confusing UX not doing it, since selections can be any size.

t2b3: Ok, I revised it and put the information into the file header. Is this ok?

src/cell.h
```diff
@@ -300,20 +300,27 @@ struct Cell {
...
-            dos.Write8(TS_GRID);
-            grid->Save(dos);
+            cellflags = TS_GRID;
+            if(this == ocs) cellflags |= TS_SELECTION_MASK;
+            dos.Write8(cellflags);
+            grid->Save(dos, ocs);
         } else {
-            dos.Write8(TS_NEITHER);
+            cellflags = TS_NEITHER;
+            if(this == ocs) cellflags |= TS_SELECTION_MASK;
```
aardappel, line 322: why is this done for each branch of `if` ? please move all code thats the same out of the `if`

src/document.h
```diff
@@ -159,7 +170,9 @@ struct Document {
 
     const wxChar *SaveDB(bool *success, bool istempfile = false, int page = -1) {
         if (filename.empty()) return _(L"Save cancelled.");
-
+        uint xs = selected.xs;
```
aardappel, line 173: if variables are used once, just remove them, and use their values below

src/cell.h
t2b3: Thanks for your feedback. I revised this. Is it ok now?

src/document.h
t2b3: Ok, I corrected this. Thanks!

aardappel:
Looks good!

t2b3:
Thanks for your review and your insights! Have a nice day over there in America!

PR #345, opened 2023-03-03
Read xs and ys value properly from the file

t2b3:
This fixes an incorrect reading of xs and ys values leading to a SEGFAULT because of garbage values supplied to DrawSelect otherwise.

### DIFF
--- a/src/document.h
+++ b/src/document.h
@@ -137,7 +137,7 @@ struct Document {
 
     uint Background() { return rootgrid ? rootgrid->cellcolor : 0xFFFFFF; }
 
-    void InitWith(Cell *r, wxString filename, Cell *ics, uint xs, uint ys) {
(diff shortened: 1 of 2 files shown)


aardappel:
Ok, that means you did not test it before submitting the last PR? Please make sure to always test locally, I assume that you, since we have no CI for interactive tests.

t2b3:
> Ok, that means you did not test it before submitting the last PR? Please make sure to always test locally, I assume that you, since we have no CI for interactive tests.

I did test locally in my local git repository and it worked for debug builds without that issue surfacing but when I installed it via the package manager on my local system (as "deployment") I noticed the flaw and I was a little bit shocked. Better double-check in the future (especially because it even worked "sometimes" with that flaw)!

PR #346, opened 2023-03-05
Darken non-matching cells in search

t2b3:
This is the second iteration of a Pull Request for darkening non-matching cells in search.

- There is only one string search left (`IsInSearch()`). In order to achieve this, there is no extra check in the cell to color it red in case of match. Darkening the surrounding non-matching cells should be highlighting enough and this change bringes the advantage that the original cell color is not changed (e.g. tags, other text colors)

### DIFF
--- a/src/document.h
+++ b/src/document.h
@@ -1065,7 +1065,7 @@ struct Document {
             case A_CASESENSITIVESEARCH: {
                 sys->casesensitivesearch = !(sys->casesensitivesearch);
                 sys->cfg->Write(L"casesensitivesearch", sys->casesensitivesearch);
-                sys->frame->search(sys->searchstring);
+                this->Refresh();
                 return nullptr;
             }
 
--- a/src/myframe.h
+++ b/src/myframe.h
@@ -1036,10 +1036,8 @@ struct MyFrame : wxFrame {
     }
 
     void OnSearch(wxCommandEvent &ce) {
-        return search(ce.GetString());
-    }
-
-    void search(wxString searchstring) {
+        wxString searchstring = ce.GetString();
+        sys->darkennonmatchingcells = searchstring.Len() != 0;
         sys->searchstring = searchstring;
         Document *doc = GetCurTab()->doc;
         doc->selected.g = nullptr;
--- a/src/system.h
+++ b/src/system.h
@@ -89,6 +89,7 @@ struct System {
     bool fswatch;
     bool autohtmlexport;
     bool casesensitivesearch;
+    bool darkennonmatchingcells;
 
     int sortcolumn, sortxs, sortdescending;
 
@@ -141,6 +142,7 @@ struct System {
           fswatch(true),
           autohtmlexport(false),
           casesensitivesearch(true),
+          darkennonmatchingcells(false),
           insidefiledialog(false) {
         static const wxDash glpattern[] = {1, 3};
         pen_gridlines.SetDashes(2, glpattern);
@@ -163,7 +165,6 @@ struct System {
         cfg->Read(L"fswatch", &fswatch, fswatch);
         cfg->Read(L"autohtmlexport", &autohtmlexport, autohtmlexport);
         cfg->Read(L"casesensitivesearch", &casesensitivesearch, casesensitivesearch);
-
         cfg->Read(L"defaultfontsize", &g_deftextsize, g_deftextsize);
 
         // fsw.Connect(wxID_ANY, wxID_ANY, wxEVT_FSWATCHER,
--- a/src/text.h
+++ b/src/text.h
@@ -177,6 +177,7 @@ struct Text {
         else
             return sys->searchstring.Len() && t.Lower().Find(sys->searchstring.Lower()) >= 0;
     }
+    
     int Render(Document *doc, int bx, int by, int depth, wxDC &dc, int &leftoffset,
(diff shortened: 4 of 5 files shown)


src/myframe.h
```diff
@@ -1036,10 +1036,8 @@ struct MyFrame : wxFrame {
     }
 
     void OnSearch(wxCommandEvent &ce) {
-        return search(ce.GetString());
-    }
-
-    void search(wxString searchstring) {
```
t2b3, line 1042: Partially revert the change before because there is no need to split the function. Instead of trigger a new search in case case sensitivity is toggled (see above), just refresh the document.

src/document.h
```diff
@@ -1065,7 +1065,7 @@ struct Document {
             case A_CASESENSITIVESEARCH: {
                 sys->casesensitivesearch = !(sys->casesensitivesearch);
                 sys->cfg->Write(L"casesensitivesearch", sys->casesensitivesearch);
-                sys->frame->search(sys->searchstring);
+                this->Refresh();
```
t2b3, line 1068: In case the case sensitivity of the search is toggled, just refresh the document so if one wants to jump to the next matching cells (i.e. the now matched cells) does not have to start from the beginning of the document (F3).

src/text.h
```diff
@@ -197,12 +198,9 @@ struct Text {
         leftoffset = h;
         int i = 0;
         int lines = 0;
-        bool searchfound = IsInSearch();
```
t2b3, line 200: Give up this string search and the possibility to color the text red.

src/system.h
```diff
@@ -89,6 +89,7 @@ struct System {
     bool fswatch;
     bool autohtmlexport;
     bool casesensitivesearch;
+    bool _darkennonmatchingcells;
```
aardappel, line 92: why does this have an `_` ?

aardappel (commented):
Ok, this looks a lot better! can you post a screenshot of what this looks like?

src/myframe.h
```diff
@@ -1036,10 +1036,8 @@ struct MyFrame : wxFrame {
     }
 
     void OnSearch(wxCommandEvent &ce) {
-        return search(ce.GetString());
-    }
-
-    void search(wxString searchstring) {
+        wxString searchstring = ce.GetString();
+        sys->_darkennonmatchingcells = (searchstring.Len() != 0);
```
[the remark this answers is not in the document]
t2b3: Thanks! Corrected

src/system.h
t2b3: This was indeed a remnant from the previous proposal. Corrected it.

t2b3:
How it looks like when search term is entered:
image

aardappel:
Cool, thanks!

aardappel:
@t2b3 this functionality is cool, the downside is that we now need easy ways to clear/cancel the search.

Before, you could search, and you could mix looking at search hits and editing them.

Now, because the sheet is darkened, you need to explicitly cancel the "darken" mode, which is unfortunate.
We can go two ways with this:
1) go back to the old idea of highlighting search items and remove darkening.
2) make it really easy to get away from darking, at least:
   - An X icon to the right of the search box to very easily cancel the search with the mouse.
   - When pressing Esc while a search is active it cancels the search.

Would you work on either one of those solutions?

t2b3:
Hello Wouter

Thanks for sharing your thoughts on this!
I see your point and I would propose to use the Search Control provided by WxWidgets, see https://docs.wxwidgets.org/latest/classwx_search_ctrl.html

This has to the advantage of providing a way to empty the value in the box by clicking on the x icon. What do you think about it?

Regards,
Tobias


Gesendet von Outlook für iOS<https://aka.ms/o0ukef>
________________________________
Von: Wouter van Oortmerssen ***@***.***>
Gesendet: Monday, April 10, 2023 10:12:19 PM
An: aardappel/treesheets ***@***.***>
Cc: Tobias Predel ***@***.***>; Mention ***@***.***>
Betreff: Re: [aardappel/treesheets] Darken non-matching cells in search (PR #346)


@t2b3<https://github.com/t2b3> this functionality is cool, the downside is that we now need easy ways to clear/cancel the search.

Before, you could search, and you could mix looking at search hits and editing them.

Now, because the sheet is darkened, you need to explicitly cancel the "darken" mode, which is unfortunate.
We can go two ways with this:

  1.  go back to the old idea of highlighting search items and remove darkening.
  2.  make it really easy to get away from darking, at least:
     *   An X icon to the right of the search box to very easily cancel the search with the mouse.
     *   When pressing Esc while a search is active it cancels the search.

Would you work on either one of those solutions?

aardappel:
Yup that search control sounds great!
I think we also want the Esc to cancel from keyboard, though.

t2b3:
I have a PoC in https://github.com/t2b3/treesheets/tree/search
But it is working only on Linux
I tried multiple times on Windows with different ways, but
Enter key or TEXT_ENTER event is not fetched in the Search Control
and the Search Control always loses focus making it impractical to bind
escape key press to it.


Am Di., 11. Apr. 2023 um 21:06 Uhr schrieb Wouter van Oortmerssen <
***@***.***>:

aardappel:
Well, instead of `wxSearchCtrl`, start by adding an `X` button right behind the search box that works the same on all OSes that just does a clear search?

PR #347, opened 2023-03-05
Jump to cell text end when entering the cell via F2

t2b3:
Instead of selecting the whole text to be overwritten when the cell is entered, set cursor to the text end. It might be worth a discussion which use case is more prominent: Overwriting cell text or appending to cell text?

### DIFF
--- a/src/document.h
+++ b/src/document.h
@@ -1370,12 +1370,13 @@ struct Document {
                 return nullptr;
 
             case A_ENTERCELL:
+            case A_ENTERCELL_JUMPTOEND:
             case A_PROGRESSCELL: {
                 if (!(c = selected.ThinExpand(this))) return OneCell();
                 if (selected.TextEdit()) {
                     selected.Cursor(this, (k==A_ENTERCELL ? A_DOWN : A_RIGHT), false, false, dc, true);
                 } else {
-                    selected.EnterEdit(this, 0, (int)c->text.t.Len());
+                    selected.EnterEdit(this, (k == A_ENTERCELL_JUMPTOEND) ? (int)c->text.t.Len() : 0, (int)c->text.t.Len());
                     DrawSelectMove(dc, selected, true);
                 }
                 return nullptr;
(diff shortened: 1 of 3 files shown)


jn64:
Currently there are two shortcuts for entering text edit: F2 and Return. Maybe they can be split to the 2 different behaviours.

I believe on some common spreadsheet programs, F2 puts the cursor at text end. So I suggest using F2 for this in treesheets as well. Return can stay as the current behaviour.

aardappel:
I would agree that wanting to add to the end is likely more common than replacing all existing content, but the latter is what people expect from this kind of UI, and we shouldn't change that lightly.

I simply got used to press Enter followed by Right to insert at the end. The advantage of selecting all is that inserting both at the start and the end is really quick this way, as opposed to if its meaning were changed to place the cursor at the end, then selecting all is not so easy anymore, since Ctrl+A works at cell level currently.. hold Ctrl+Shift+Left I guess.

Agree with @jn64 that we could change F2 to do this since that is probably less used.

georgeraraujo:
> Currently there are two shortcuts for entering text edit: F2 and Return. Maybe they can be split to the 2 different behaviours.
> 
> I believe on some common spreadsheet programs, F2 puts the cursor at text end. So I suggest using F2 for this in treesheets as well. Return can stay as the current behaviour.

I just tested and in both Excel for Office 365 and LibreOffice Calc 7.4.0.3 on Windows 10, F2 puts the cursor at the text end, and Return skips to the cell below.

t2b3:
Thank you very much for your feedback, making the proposals much better! I agree with you that we can split the behaviour:
When having pressed Return to enter the cell, select the whole cell text as it is done
When having pressed F2 to enter the cell, jump the cursor to the end of the cell text as also other spreadsheet programs do.
I have revised the Pull Request accordingly. What do you think of it?
I am looking forward to hearing from you!
Kind regards,
Tobias

aardappel:
LGTM!

PR #348, opened 2023-03-05
Proposals for improving the search

t2b3:
Proposals:
- Selection should be kept when search is active (e.g. when the user changes the search term or has mistyped the search term, it may be more convenient to keep the selection)
- When pressing Enter in the Search text control with no search term, give back focus to the Canvas window.


aardappel:
not sure why the selection was deleted before, that may have something to do with what happens if the first search result is out of view, or folded/minimized etc, worth trying if that all still works

Issue #328, opened 2023-01-02
Translation churn


aardappel:
In recent months, there have been many contributions to the translation, which generally is very welcome and awesome.

On the downside, over half of recent PRs have all been small translation fixes. This costs time for me to check and produces a lot of churn (commits/releases..).

We can't have the situation where an actual PR adds 1 string to the code for a feature, and immediately we get 5 language PRs adding that string.

I also don't think it is necessary. TreeSheets is a program with a medium amount of popularity, but the amount of users of 1 language that are going to get an updated exe and then use this new feature and get to see an english string is really rather miniscule.

There are various ways we could solve this, but I propose a simple one: lets update translations for a language at most every quarter (3 months).

What do you all think?

Tagging recent translation authors: @t2b3 @georgeraraujo @Philippe-devel-github @albanobattistella @Rasphino

t2b3:
Hello Wouter

Thanks for explaining your issue with the amount of translation PRs.
I can understand your point of view and I would have no problem to collect translation updates quarterly.

Regards
Tobias

georgeraraujo:
> There are various ways we could solve this, but I propose a simple one: **lets update translations for a language at most every quarter (3 months)**.
> 
> What do you all think?

Sounds good to me. That would also be an opportunity to do a tagged, versioned release that package maintainers could pick up.

aardappel:
We have an automaticly tagged versioned release system now that people can use.. I don't think we will do manual versions on top of that / conflicting with that.

georgeraraujo:
> We have an automaticly tagged versioned release system now that people can use.. I don't think we will do manual versions on top of that / conflicting with that.

I understand. It's just that the [Debian package](https://packages.debian.org/sid/treesheets), even in sid (unstable), is ancient -- 1.0.2 was released in Aug 2, 2019. I suppose a, say, 1.0.3 could be easier for package maintainers to pick up. But, again, I understand.

aardappel:
They can pick up any of the recent tagged versions if they want.

georgeraraujo:
Ok.

As for submitting translation updates, as I said earlier, I’m fine with doing so at most every quarter.

PR #354, opened 2023-04-02
pt_BR: update translation

georgeraraujo:


aardappel:
I don't think we should be updating these files each time the code changes. There has to be a better way. Maybe we can generate the binaries during CI or something.

georgeraraujo:
I’ve done as discussed in #328.

PR #355, opened 2023-04-04
Adapt tutorial file for new search behaviour

t2b3:
Matching cell text will not be red, but the matching cells are highlighted, i.e. the non-matching cells are darkened.


aardappel:
Tsk, github doesn't know how to diff TS files? :P

t2b3:
> Tsk, github doesn't know how to diff TS files? :P

How dare they?  :grinning: Maybe it is going to change... who knows? ;-)

PR #356, opened 2023-04-05
Jump to the selection after first paint on canvas

t2b3:
Use case: If the user opens the TreeSheets document, TreeSheets will jump to the selection.

Benefit: The user can continue right away at the place where the user stopped before.

Caveat: Tested on Windows and Linux, but only works on Linux so far (reason unknown).


aardappel:
If it only works on Linux, that probably means its doing something "unreliable", which means it probably should not be merged :)

PR #357, opened 2023-04-06
Use wxPaintDC instead of wxBufferPaintDC on Windows

t2b3:
I tested on Windows 10. There seems to be no need for `wxBufferPaintDC` (any more).


aardappel:
Last I checked the scrolling felt much worse without it..

aardappel:
Yup, just tested on Windows 10 also, and its a flickering mess with your change. You can't see a difference?

t2b3:
Thanks for having checked this PR. I noticed the differences now but they were more subtle in my case. I withdraw my PR.

PR #358, opened 2023-04-06
Jump to selection after first paint on Linux

t2b3:
I tested under Linux and it works. I would like to make this a Linux-exclusive feature for now.


aardappel:
It still means it is likely doing something unintended, and may break on Linux too. I would suggest not doing it.

PR #359, opened 2023-04-07
Scroll to selection on first draw

t2b3:
I tried it again and I found a working solution for both Linux and Windows


aardappel:
Ok, lets try this out then :)

t2b3:
> Ok, lets try this out then :)

Thanks for the merge! :+1:

PR #361, opened 2023-04-07
Add Doxygen support

t2b3:
Documentation can be automatically generated by issuing `make doc`


aardappel:
We don't have doxygen source code comments nor do we have doxygen style markdown docs.. what is the use of this?

And if there is a use, it should be highly optional, i.e. not require doxygen to be installed.

t2b3:
> We don't have doxygen source code comments nor do we have doxygen style markdown docs.. what is the use of this?
> 
> And if there is a use, it should be highly optional, i.e. not require doxygen to be installed.

Thanks for your feedback!
I think it is useful to have the collaboration and inheritance diagrams in order to understand how the classes are linked to each other. I made this feature optional. I tested it with doxygen installed and not installed.

aardappel:
> I think it is useful to have the collaboration and inheritance diagrams in order to understand how the classes are linked to each other.

TS is not object oriented (hardly uses inheritance) and is very small and simple code, so I think this is not particularly useful to anyone else.. I suggest just running doxygen locally.

PR #363, opened 2023-04-08
Clarify file format specification

t2b3:
It is not any of the cells in the selection, but only the first cell within the selected that is marked with the highest bit.


t2b3:
Thanks for the merge!

PR #366, opened 2023-04-09
Remove obsolete function MakeInternallyScaled()

t2b3:
As wxWidgets handles the correct drawing depending on the set DPI by using device-independent pixels, the function MakeInternallyScaled() becomes obsolete. This commit removes the function.


aardappel:
Haven't tried this on Mac lately, but if its unused, I guess that's ok!

t2b3:
Thanks for the merge!

PR #367, opened 2023-04-10
Remove bm_orig

t2b3:
* Fill image_data and image_type automatically when Bitmap is provided
* Generate bm_display out of the image_data vector
* Convert directly when user switches between PNG and JPEG format
* Make BitmapScale() not use bm_orig
* Only place PNG image in clipboard (on Linux, on other platforms place Bitmap)


t2b3:
This commit has been tested on Linux and Windows.
After purge and having a TreeSheets open with many images, the memory consumption was lowered from 88.8 MiB (before this commit) to 76.5 MiB (after this commit), thus making a reduce of memory consumption in idle state of about 14%.

aardappel:
You saw the comment saying `this is WAY faster than recompressing (~30x on image heavy files)` right? that code was introduced when I got a test file from the user that enormous amounts of images in it.

Please create a TS file with 100 large pngs and jpegs in it, and measure saving time before and after your change if you wish to remove that feature. Only if there's not a lot of difference on Win/Lin we can do it.

14% memory usage is a good savings, but its nothing against 30x saving speed.

t2b3:
Hello Wouter

Thanks for your feedback!

I think that you might misread the code. On the contrary, the code makes it the only way to directly use the data from the vector and weites it directly to disk. To ensure that it is working the code makes sure that the vector is always filled with Image Data.

The memory saving comes on top.
Do you agree?

Regards
Tobias

Gesendet von Outlook für iOS<https://aka.ms/o0ukef>
________________________________
Von: Wouter van Oortmerssen ***@***.***>
Gesendet: Monday, April 10, 2023 5:05:54 PM
An: aardappel/treesheets ***@***.***>
Cc: Tobias Predel ***@***.***>; Author ***@***.***>
Betreff: Re: [aardappel/treesheets] Remove bm_orig (PR #367)


You saw the comment saying this is WAY faster than recompressing (~30x on image heavy files) right? that code was introduced when I got a test file from the user that enormous amounts of images in it.

Please create a TS file with 100 large pngs and jpegs in it, and measure saving time before and after your change if you wish to remove that feature. Only if there's not a lot of difference on Win/Lin we can do it.

14% memory usage is a good savings, but its nothing against 30x saving speed.

src/system.h
```diff
@@ -47,8 +53,36 @@ struct Image {
...
+        im.SaveFile(mos, bmt);
+        off_t afterimage = mos.TellO();
+        auto sz = afterimage - beforeimage;
+        wxMemoryInputStream mis(mos);
+        pidv.resize(sz);
+        mis.Read(pidv.data(), pidv.size());
+        return pidv;
+    }
+
+    wxImage ConvertVectorToImage(vector<uint8_t> pidv, wxBitmapType bmt) {
```
aardappel, line 72: you're passing a vector by value, its going to get copied.. please pass by ref

src/system.h
```diff
@@ -47,8 +53,36 @@ struct Image {
     void Purge() {
         auto seconds_passed = wxGetLocalTime() - last_display;
         if (seconds_passed > 10) bm_display = wxNullBitmap;
-        // TODO: we could even purge bm_orig for even greater memory savings if we have png_data.
     }
+
+
+    vector<uint8_t> ConvertImageToVector(const wxImage &im, wxBitmapType bmt) {
```
aardappel, line 59: Vector is confusing, Buffer would be better

src/system.h
```diff
@@ -17,12 +15,18 @@ struct Image {
 
     long last_display = 0;
 
-    Image(wxBitmap _bm, uint64_t _hash, double _sc, vector<uint8_t> &&pd, char iti)
-        : bm_orig(_bm), hash(_hash), display_scale(_sc), image_data(std::move(pd)), image_type(iti) {}
+    Image(const wxBitmap &bm, uint64_t _hash, double _sc, vector<uint8_t> &&pd, char iti)
```
aardappel, line 18: Note that the caller for this constructor is often `AddImageToList`, which got an `Image` and `vector` from loading, which then converts it to a bitmap, an you then convert that bitmap back.. this constructor should not take a bitmap.

Also you're copying `image_data` from the vector, then overwriting it again with the bitmap???? Please be sure you understand the code you modify precisely.

src/system.h
```diff
@@ -38,7 +42,9 @@ struct Image {
 
     wxBitmap &Display() {
         if (!bm_display.IsOk()) {
-            ScaleBitmap(bm_orig, 1.0 / display_scale * sys->frame->csf, bm_display);
+            wxBitmapType it = imagetypes[image_type].first;
+            wxBitmap bm = ConvertVectorToBitmap(image_data, it);
```
aardappel, line 46: So, effectively the cost at loading of converting to bitmap is delayed until here.

src/system.h
t2b3: Thanks for this remark! Solved with 1627303bcda085888f066350bb1a6732174f443d

src/system.h
t2b3: Ok, I see your point. Solved with https://github.com/aardappel/treesheets/pull/367/commits/5e99efa6393f15b52067e2f51ef26904ab3b6a51

src/system.h
t2b3: I see your point. This PR is only the half way. The other half way comes (hopefully) with https://github.com/aardappel/treesheets/pull/367/commits/aca1a28dd69aa3a4e99de341fb8e9a2f5fd0260c

t2b3:
| | PNG | JPEG |
| --- | --- | --- |
| before PR | m_i = 84.7 MiB, t_L = 298 ms | m_i = 81.3 MiB, t_L = 129 ms |
| after PR | m_ = 71.1 MiB, t_L = 267 ms | m_i = 53.0 MiB, t_L = 84 ms |

src/system.h
aardappel: PRs should be self contained and do only one thing. They should not contain unrelated changes or rely on changes in future PRs. Can you make this more reviewable? I can't know what you're all still planning to do.

aardappel:
Thanks for the benchmark numbers, that is very helpful

src/system.h
t2b3: I think that the PR should be now ready. It should do one thing: remove bm_orig with all its consequences
Is this ok for you?

aardappel:
Ok, I think it looks good now, lets try it :)

PR #369, opened 2023-04-16
Leave search when Escape key is pressed

t2b3:
This PR uses the existing workaround with menu events processed by OnMenu.
Escape key is bound to either action A_ENTERCELL or A_ENTERGRID, so these are used as triggers to clear the search control.

For now, I would recommend to not use the wxSearchCtrl because it is a native widget for wxGTK and wxGTK but not in wxMSW, leading to different behaviour on events.

### DIFF
--- a/src/document.h
+++ b/src/document.h
@@ -1080,7 +1080,7 @@ struct Document {
             }
 
             case A_SEARCHNEXT: {
-                return SearchNext(dc, true);
+                return SearchNext(dc, false);
             }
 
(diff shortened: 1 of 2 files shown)


t2b3:
I keep https://github.com/t2b3/treesheets/tree/search as a separate branch with the workaround removed.
This only works with a native search control widget and thus not on Windows.

aardappel:
Thanks!

PR #373, opened 2023-05-01
Update window when DPI is changed

t2b3:
This forces the window and its subchildren to repaint. Please refer to <https://docs.wxwidgets.org/latest/classwx_window.html> for more ddetails.

### DIFF
--- a/src/myframe.h
+++ b/src/myframe.h
@@ -1066,10 +1066,11 @@ struct MyFrame : wxFrame {
     }
 
     void OnDPIChanged(wxDPIChangedEvent &dce) {
-          if (nb) loop(i, nb->GetPageCount()) {
-                TSCanvas *p = (TSCanvas *)nb->GetPage(i);
-                p->doc->dpichanged = true;
-                }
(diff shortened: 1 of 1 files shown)


src/myframe.h
```diff
@@ -1070,6 +1070,7 @@ struct MyFrame : wxFrame {
                 TSCanvas *p = (TSCanvas *)nb->GetPage(i);
                 p->doc->dpichanged = true;
                 }
+        Update();
```
aardappel, line 1073: Fix indentation of this function?
Why does each doc have its own `dpichanged` ? Can it different for different docs? Otherwise move to System or Frame?

t2b3: Thanks for your feedback. Yes, `dpichanged` signals that the layout and the children must be reset and this is per document. Othewise you may need to be make this a global state and thus check when the last document has been successfully painted? This adds complexity somewhere elese I think.

aardappel: ok, lets leave it then!

aardappel:
Thanks!

PR #374, opened 2023-05-02
Adjust tab height of notebook in case DPI changed

t2b3:
The tab Control height does not get resized automatically so resize it manually in case the DPI changed

Please refer to <https://docs.wxwidgets.org/latest/classwx_aui_notebook.html#aa784a41316fc5bde864d90ae0a3dac8a> for more details.

### DIFF
--- a/src/mywxtools.h
+++ b/src/mywxtools.h
@@ -107,18 +107,16 @@ struct ImagePopup : wxVListBoxComboPopup {
 
 struct ImageDropdown : wxOwnerDrawnComboBox {
     // FIXME: delete these somewhere
-    Vector<wxBitmap *> bitmaps_display;
+    Vector<wxBitmap *> bitmaps;
     wxArrayString as;
     const int image_space = 22;
 
     ImageDropdown(wxWindow *parent, wxString &path) {
         wxString f = wxFindFirstFile(path + L"*.*");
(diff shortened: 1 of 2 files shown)


src/mywxtools.h
```diff
@@ -138,8 +136,10 @@ struct ImageDropdown : wxOwnerDrawnComboBox {
     }
 
     void OnDrawItem(wxDC &dc, const wxRect &rect, int item, int flags) const {
-        auto bm = bitmaps_display[item];
-        sys->ImageDraw(bm, dc, rect.x + FromDIP(3), rect.y + FromDIP(3));
+        auto bm = bitmaps[item];
+        wxBitmap sbm;
+        ScaleBitmap(*bm, FromDIP(1.0) / dd_icon_res_scale, sbm);
```
aardappel, line 141: This change seems unrelated from the tab height, please put these in separate PRs!

Also scaling a bitmap on every render seems expensive? Why not stick this bitmap initialization in a function, and re-run it if DPI changes?

t2b3: Hello Wouter,
Thanks for your reply! I opened a separated PR for the first commit.
And also a big thanks for your input on the second commit. This seems a better way to go. I will look how I implement this.
Regards
Tobias

PR #377, opened 2023-05-03
Rescale images in image dropdown in case DPI changed

t2b3:
In order to refill the vector with the bitmaps when the DPI changes, isolate the corresponding code block into a generic member function called FillBitmapVector. This commit also places the variable imagepath into the MyFrame class so that it is accessible for the function that handles the DPI change event


aardappel:
perfect :)

PR #379, opened 2023-05-06
Add toolbar icons for replace control

t2b3:
This commit adds two toolbar icons for replacing all filter/search occurences with the replacement text either in the selection or in the document.

This commit also adds a toolbar icon to clear the replacement text in the text control like in the search text control.

When all occurences are replaced in the document, the search and replacement text will be swapped so that the user still sees the replacements and can "undo" the operation by repeating it.

This commit also rearranges the code a little bit to avoid redundancies.

This commit considers the cases for case sensitive search and  for case insensitive search.


aardappel:
Screenshot of what this looks like?

Those icons don't look very much like "replace", they look more like "play" and "fast forward" :)

t2b3:
> Screenshot of what this looks like?
> 
> Those icons don't look very much like "replace", they look more like "play" and "fast forward" :)

It looks like this currently:

https://user-images.githubusercontent.com/3286756/236662613-2dde6915-8f87-49f4-963c-04169d250fa6.mp4

Yes, I could not not find any other icons in the nuvola iconset so I took these as kind of last resort.

t2b3:
> > Screenshot of what this looks like?
> > Those icons don't look very much like "replace", they look more like "play" and "fast forward" :)
> 
> It looks like this currently:
> demo_replace_ctrls.mp4
> 
> Yes, I could not not find any other icons in the nuvola iconset so I took these as kind 

Here the demonstration for the other toolbar button:
https://user-images.githubusercontent.com/3286756/236662847-41a63f91-caf2-4035-878b-c9c3d2805e20.mp4

t2b3:
Whole walkthrough of the feature:

https://user-images.githubusercontent.com/3286756/236667510-2bbd2bc1-c5d3-4600-b0ef-a0bc527710ef.mp4

t2b3:
Unfortunately Ctrl+h is superseded by Backspace on Linux

https://user-images.githubusercontent.com/3286756/236668378-b2ef7f9e-c18b-4ff9-98bb-b58030af5f0c.mp4

aardappel:
The functionality looks nice.

One unfortunate effect is the bar keeps getting longer.. I can already hear the complaints of the people on 1280 wide laptop screens coming.. :)

Well, we certainly can't have the blue same icon for both `Run` and `Replace` in the toolbar.

If you google "icon for replace", you get lots of these kinds of icons:

download

Does nuvola have anything with 2 opposing arrows? If not we'll have to just pick one from a non-nuvola set.
Not sure how to express "replace all", google doesn't show different icons for it. Maybe we'll make replace all red to indicate a higher level of danger.

Ah found some from nuvola:

image
image

It's a bit low res, but better than nothing..

src/document.h
```diff
@@ -1117,12 +1111,31 @@ struct Document {
...
             case A_REPLACEALL: {
                 if (!sys->searchstring.Len()) return _(L"No search.");
-                rootgrid->AddUndo(this);  // expensive?
-                rootgrid->FindReplaceAll(sys->frame->replaces->GetValue());
-                rootgrid->ResetChildren();
-                Refresh();
+                if (k == A_REPLACEALL) {
+                    wxString filter = sys->frame->filter->GetValue();
+                    wxString replaces = sys->frame->replaces->GetValue();
+                    rootgrid->AddUndo(this);  // expensive?
```
aardappel, line 1121: Yes that is going to copy the whole document. Since there is a cap on the amount of memory undoes can hold, the bad effect of this is that it may clear the rest of the undo history.

The alternative is a single undo per replaced cell. That is likely ok if its just a few replacements but annoying for a large amount..

Better yet would be complicated: first find where all the replacements will take place, then create an undo only for that area, then do the replacement

src/document.h
```diff
@@ -1117,12 +1111,31 @@ struct Document {
...
+                    rootgrid->FindReplaceAll(replaces);
+                    sys->frame->filter->SetValue(replaces);
```
aardappel, line 1123: why is it needed to be set again

src/document.h
```diff
@@ -1117,12 +1111,31 @@ struct Document {
...
+                } else {
+                    selected.g->ReplaceStr(this, sys->frame->replaces->GetValue(), selected);
+                    if (k == A_REPLACEONCEJ) return SearchNext(dc, true);
+                }
+                return _(L"Text has been replaced.");
+            }
+
+             case A_CLEARREPLACE:
+             case A_CLEARSEARCH: {
+                if (k == A_CLEARREPLACE) sys->frame->replaces->Clear();
```
aardappel, line 1136: This can be 2 cases

src/text.h
```diff
@@ -410,12 +410,22 @@ struct Text {
...
+        if(sys->casesensitivesearch) {
+            for (int i = 0, j; (j = t.Mid(i).Find(sys->searchstring)) >= 0;) {
             // does this need WasEdited()?
             i += j;
             t.Remove(i, sys->searchstring.Len());
             t.insert(i, str);
             i += str.Len();
+            }
+        } else {
+            for (int i = 0, j; (j = t.Mid(i).Lower().Find(sys->searchstring.Lower())) >= 0;) {
```
aardappel, line 422: the only difference between these two loops is whether the search string is lowered? why don't you do that out of the loop, or at the start of search? These are not cheap operations

src/document.h
t2b3: Thanks! I corrected this.

src/text.h
t2b3: Thanks for the hints! I found another way to these operations before the loop. Thanks :-)

t2b3:
> The functionality looks nice.
> 
> One unfortunate effect is the bar keeps getting longer.. I can already hear the complaints of the people on 1280 wide laptop screens coming.. :)
> 
> Well, we certainly can't have the blue same icon for both `Run` and `Replace` in the toolbar.
> 
> If you google "icon for replace", you get lots of these kinds of icons:
> 
> download
> 
> Does nuvola have anything with 2 opposing arrows? If not we'll have to just pick one from a non-nuvola set. Not sure how to express "replace all", google doesn't show different icons for it. Maybe we'll make replace all red to indicate a higher level of danger.
> 
> Ah found some from nuvola:
> 
> image image
> 
> It's a bit low res, but better than nothing..

Thanks for looking up for suitable icons. I took two from the nuvola set. What do you think of these?
image

t2b3:
Hi Wouter

I revised my commit now and tries to incorporate your remarks. What do you think of it now?
Happy to have your review!

Regards and nice greetings to USA from Switzerland
Tobias

src/document.h
```diff
@@ -1117,12 +1111,36 @@ struct Document {
...
             case A_REPLACEALL: {
                 if (!sys->searchstring.Len()) return _(L"No search.");
-                rootgrid->AddUndo(this);  // expensive?
-                rootgrid->FindReplaceAll(sys->frame->replaces->GetValue());
-                rootgrid->ResetChildren();
-                Refresh();
+                if (k == A_REPLACEALL) {
+                    wxString filter = sys->frame->filter->GetValue();
+                    wxString replaces = sys->frame->replaces->GetValue();
+                    loopallcells(c) if(c->text.IsInSearch()) c->AddUndo(this);
```
aardappel, line 1121: Hmm this is also potentially a lot.. we may need something better for this, but I guess that can be a future change..

aardappel, line 1121: This loops over all cells in `c` not `rootgrid` ???

src/text.h
```diff
@@ -410,10 +410,12 @@ struct Text {
     }
 
     void ReplaceStr(const wxString &str) {
-        for (int i = 0, j; (j = t.Mid(i).Lower().Find(sys->searchstring)) >= 0;) {
+        wxString text = sys->casesensitivesearch ? t : t.Lower();
```
aardappel, line 413: This copies the entire string even in the case `Lower` does not need to be called. Also, the search string can be converted just once, at the start of the search, rather than for reach cell.

aardappel:
The replace all icon is good!

t2b3:
Hi Wouter

I took your remarks and have adjusted the PR.
I would like to kindly ask you to take a look over it.

Kind regards,
Tobias

src/document.h
aardappel: this comment has not been addressed

src/document.h
aardappel: this comment has not been addressed

src/document.h
```diff
@@ -1117,12 +1115,36 @@ struct Document {
...
+                    rootgrid->FindReplaceAll(replaces);
+                    sys->frame->filter->SetValue(replaces);
+                    sys->frame->replaces->SetValue(filter);
+                    rootgrid->ResetChildren();
+                    Refresh();
+                } else {
+                    loopallcellssel(c, true) if(c->text.IsInSearch()) c->AddUndo(this);
```
aardappel, line 1132: this seems to repeat code from the other if branch

src/text.h
```diff
@@ -410,12 +410,22 @@ struct Text {
...
+        if(sys->casesensitivesearch) {
+            for (int i = 0, j; (j = t.Mid(i).Find(sys->searchstring)) >= 0;) {
+                // does this need WasEdited()?
+                i += j;
+                t.Remove(i, sys->searchstring.Len());
+                t.insert(i, str);
+                i += str.Len();
+            }
+        } else {
+            for (int i = 0, j; (j = t.Lower().Mid(i).Find(sys->searchstring)) >= 0;) {
```
aardappel, line 422: lower `t` just once? share the loop?

t2b3:
Hello Wouter

Now I would try for the third round :-)

Regards
Tobias

src/document.h
aardappel: pull `(sys->casesensitivesearch) ? (wxString) wxEmptyString : replaces.Lower()` out of the if

aardappel:
LGTM, thanks!

t2b3:
> LGTM, thanks!

Thanks for your reviews!

PR #381, opened 2023-05-13
Update Translation template file and German translation

t2b3:
Hello Wouter

I think now is a good time to consolidate the translation.
What do you think?

Regards
Tobias


TS/translations/ts.pot
```diff
@@ -3,13 +3,13 @@
 # This file is distributed under the same license as the PACKAGE package.
 # FIRST AUTHOR <EMAIL@ADDRESS>, YEAR.
 #
-#: document.h:232
```
aardappel, line 6: yeah this file changes all the time.. its just in the comments, I think it doesn't need updating as much

aardappel:
Thanks!

PR #384, opened 2023-05-15
Ditch location comment

t2b3:
This PR deletes the location comments in the template file, in all message catalogs (text and binary).
It also adjusts the command to generate the template file that the location comments are no longer added.


aardappel:
Ahh awesome.. this will reduce churn :)

t2b3:
> Ahh awesome.. this will reduce churn :)
You are very welcome!

PR #388, opened 2023-05-17
Copy selection as Bitmap to clipboard

t2b3:
This adds the rendered subbitmap of the selection to the clipboard data composite object.

This enables to copy the bitmap to programs dealing with bitmaps, e.g. image manipulation programs or journals.

### DIFF
--- a/src/document.h
+++ b/src/document.h
@@ -708,6 +708,22 @@ struct Document {
         return ExportFile(fn, k, true);
     }
 
+    wxBitmap GetBitmap() {
+        maxx = layoutxs;
+        maxy = layoutys;
+        originx = originy = 0;
+        wxBitmap bm(maxx, maxy, 24);
+        wxMemoryDC mdc(bm);
+        DrawRectangle(mdc, Background(), 0, 0, maxx, maxy);
+        Render(mdc);
+        return bm;
+    }
+
+    wxBitmap GetSubBitmap(Selection &s) {
+        wxRect r = s.g->GetRect(this, s, true);
+        return GetBitmap().GetSubBitmap(r);
+    }
+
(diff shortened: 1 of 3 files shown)


aardappel:
Hmm.. and this only applies to application that do not understand text or html?

Seems a bit of a niche use case and a relatively heavy operation to do on each copy? Maybe better as an explicit copy as bitmap?

t2b3:
Hi Wouter
Thanks for your feedback. After having rethought this a bit, I agree with you. The best solution is to keep this rather heavy operation in an extra menu item (and the user is aware of the option). I will deal with the adjustments tomorrow then.
Regards,
Tobias

t2b3:
I am good to go from my side.

src/document.h
```diff
@@ -708,6 +708,22 @@ struct Document {
...
+        wxBitmap bm(maxx, maxy, 24);
+        wxMemoryDC mdc(bm);
+        DrawRectangle(mdc, Background(), 0, 0, maxx, maxy);
+        Render(mdc);
+        return bm;
+    }
+
+    wxBitmap GetSubBitmap(Selection &s) {
+        wxRect r = s.g->GetRect(this, s, true);
+        return GetBitmap().GetSubBitmap(r);
```
aardappel, line 724: Ok, this is not great, since you're rendering the whole view then only copying part of it.. would be better to modify GetBitmap to work correctly both with the view root and a selection.. but if its more complicated, I guess this only happens when you explicitly use copy as bitmap, I guess its ok for now.

aardappel:
Thanks!

src/document.h
t2b3: Thanks for your reply! I totally see your point, maybe I will find a way to have a better way. Render() seems to have something in document.h:533.

PR #395, opened 2023-05-20
Use std::array instead of C-style array

t2b3:
clang-tidy recommends to use std::array instead of C-style arrays


aardappel:
Hmm this is debatable if it makes things better.. but if it helps with keeping clang-tidy happy..

PR #398, opened 2023-05-20
Make imagetypes const

t2b3:
Operator[] will not work with const std::map, so use at function instead.

### DIFF
--- a/src/document.h
+++ b/src/document.h
@@ -879,13 +879,13 @@ struct Document {
                 } else {
                     // Always convert to PNG file format because wxWidgets has trouble dealing with other 
                     // image file formats in clipboard (especially for pasting).
-                    wxImage imi = sys->ConvertBufferToWxImage(im->image_data, imagetypes[im->image_type].first);
+                    wxImage imi = sys->ConvertBufferToWxImage(im->image_data, imagetypes.at(im->image_type).first);
                     vector<uint8_t> idv = sys->ConvertWxImageToBuffer(imi, wxBITMAP_TYPE_PNG);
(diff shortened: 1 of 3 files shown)


aardappel:
Yes, `[]` creates the item if it doesn't exist.

src/system.h
```diff
@@ -297,7 +297,7 @@ struct System {
                     case 'I':
                     case 'J': {
                         char iti = *buf;
-                        wxBitmapType imt = imagetypes[iti].first;
+                        wxBitmapType imt = imagetypes.at(iti).first;
```
aardappel, line 300: `at` should only be used if you're 100% sure the value is correct, since it will throw an exception. Here, `find` would be more appropriate.

src/document.h
```diff
@@ -879,13 +879,13 @@ struct Document {
                 } else {
                     // Always convert to PNG file format because wxWidgets has trouble dealing with other 
                     // image file formats in clipboard (especially for pasting).
-                    wxImage imi = sys->ConvertBufferToWxImage(im->image_data, imagetypes[im->image_type].first);
```
t2b3, line 882: We do the check whether the `image_type` really exists above in line 874.

t2b3:
If you run in a merge issue, please let me now and I will rebase :-)

src/system.h
t2b3: Thanks for your feedback. I changed it accordingly.

aardappel:
Yes this needs a rebase :)

t2b3:
Thanks!

PR #400, opened 2023-05-20
Make celltextcolors non static

t2b3:
As it is outside of a class, it does not seem to be necessary to declare celltextcolors as static.


aardappel:
`static` outside a class means its not visible outside of the compilation unit, which is still useful / preferred. It makes for linking that is faster and less likely to clash with other symbols.

PR #401, opened 2023-05-21
Make variables in main.cpp static

t2b3:
They shall not be visible outside the compilation unit.

This is the opposite direction of https://github.com/aardappel/treesheets/pull/400 that was wrong.


aardappel:
ok, if this gets rid of the warning, then no need to move them into `System` I guess

PR #402, opened 2023-05-21
Mark strings for translation & update pt_BR

georgeraraujo:


src/myapp.h
```diff
@@ -146,7 +146,7 @@ struct MyApp : wxApp {
 
         auto serr = ScriptInit(frame->GetDataPath("scripts/"));
         if (!serr.empty()) {
-            wxLogFatalError(L"Script system could not initialize: %s", serr);
+            wxLogFatalError(wxString::Format(_(L"Script system could not initialize: %s"), serr));
```
aardappel, line 149: Let's not include this file, errors from here would be highly technical errors that pretty much never happen, they don't need to be localized.
Generally, only things in the graphical UI should be localized, not logging messages.

aardappel:
@t2b3 who recently changed how the localization is set up

src/genpot.bat
```diff
@@ -1 +1 @@
-xgettext --keyword=_ --sort-output --no-location -o ../TS/translations/ts.pot document.h myapp.h myframe.h system.h
```
aardappel: revert?

src/myapp.h
```diff
@@ -146,7 +146,7 @@ struct MyApp : wxApp {
 
         auto serr = ScriptInit(frame->GetDataPath("scripts/"));
         if (!serr.empty()) {
-            wxLogFatalError(wxString::Format(_(L"Script system could not initialize: %s"), serr));
```
aardappel: revert?

PR #403, opened 2023-05-21
Make all global variables const that are not changed

t2b3:
- Rename variable name MAX_LAUNCHES so it fits to the other variables

- Rearrange the order of variables for better readibility (separate the const from the non-const variables)


t2b3:
This might go against the comments, but maybe leave it for now?

aardappel:
This is fine.

PR #408, opened 2023-05-22
Add German translation for new strings

t2b3:
Thanks to @georgeraraujo for the template update.


t2b3:
Maybe we should add something like https://cmake.org/cmake/help/latest/module/FindGettext.html to the CMake file to automatically compile the message catalogs?

aardappel:
Having that only on 1 platform doesn't sound that useful..

PR #411, opened 2023-05-23
Use Selection constructor with default arguments

t2b3:
This fixes the case where the member variables are not properly initialized in case of Selection().

This manifested in the copied HTML string when only one cell was selected because s.x was not properly initialized (`<tr>` tag was missing).


aardappel:
this is nicer, sure, but I don't understand your description? how could `Selection()` cause `x` to not be initialized? `memset` would always set it to 0?

PR #420, opened 2023-05-27
Use vector directly for hash computation

t2b3:
When hash the complete dataset, do not add extra
type conversation to string_view


aardappel:
Nice!

PR #423, opened 2023-05-29
Explicitly prepend std:: to STL vector

t2b3:
Avoid confusion with Vector


aardappel:
The reason why `vector` was possible is because somewhere in the code it probably says `using namespace std`. It would be good to remove that, so you can enforce all uses of these types consistently use `std::`.

Note that there's also such uses in the Lobster code where you probably don't want to remove it, not only is it a lot of work but its copied from another project.

t2b3:
I am not quite sure about the benefit or the cost-benefit-ratio of this. I close this for now.

aardappel:
either way, be consistent: use it always, or use it never :)

PR #428, opened 2023-06-01
Introduce ThreadPool and use it for Display()

t2b3:
Rendering the image files to bitmaps is a rather heavy task
that consumes a lot of CPU time (relatively).

Instead of rendering the images one by one,
do this in parallel with threads using the available CPU cores.

The threads are managed by ThreadPool that is introduced in this commit
as the base.

### DIFF
--- a/src/document.h
+++ b/src/document.h
@@ -879,13 +879,13 @@ struct Document {
                 } else {
                     // Always convert to PNG file format because wxWidgets has trouble dealing with other 
                     // image file formats in clipboard (especially for pasting).
-                    wxImage imi = sys->ConvertBufferToWxImage(im->image_data, imagetypes.at(im->image_type).first);
-                    vector<uint8_t> idv = sys->ConvertWxImageToBuffer(imi, wxBITMAP_TYPE_PNG);
+                    wxImage imi = ConvertBufferToWxImage(im->image_data, imagetypes.at(im->image_type).first);
+                    vector<uint8_t> idv = ConvertWxImageToBuffer(imi, wxBITMAP_TYPE_PNG);
                     image->SetData(idv.size(), idv.data());
                 }
                 wxTheClipboard->SetData(image);
                 #else
-                wxBitmap bm = sys->ConvertBufferToWxBitmap(im->image_data, imagetypes.at(im->image_type).first);
+                wxBitmap bm = ConvertBufferToWxBitmap(im->image_data, imagetypes.at(im->image_type).first);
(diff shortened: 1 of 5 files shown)


src/threadpool.h
```diff
@@ -0,0 +1,92 @@
+inline void THROW_OR_ABORT(const std::string &s) {
```
aardappel, line 1: remove this function, turn the error below into an assert

src/document.h
```diff
@@ -545,6 +545,16 @@ struct Document {
         dc.SetBackground(wxBrush(wxColor(Background())));
         dc.Clear();
         if (!rootgrid) return;
+        loopv(i, sys->imagelist) sys->imagelist[i]->trefc = 0;
+        rootgrid->ImageRefCount();
```
aardappel, line 549: Why is this happening on every draw? better to do it at the end of loading

t2b3: Ah sorry. I changed that.

src/threadpool.h
t2b3: Ok I did that.

src/system.h
```diff
@@ -406,6 +406,16 @@ struct System {
 
     done:
 
+        loopv(i, sys->imagelist) sys->imagelist[i]->trefc = 0;
+        doc->rootgrid->ImageRefCount();
+
+        ThreadPool pool(std::thread::hardware_concurrency());   
+        loopv(i, sys->imagelist) {
+            pool.enqueue([](Image *img) {
+                if(img->trefc) img->Display();
```
aardappel, line 415: It be good to put a comment inside `Display` that this runs in multiple threads.. i.e. if anyone ever modifies display, they should not be touching any global resources, and we're assuming that the wx functions its calling are thread safe (do we know?)

t2b3: Ok I added a comment.

src/system.h
```diff
@@ -41,6 +41,9 @@ struct Image {
     }
 
     wxBitmap &Display() {
+        // This might run in multiple threads in parallel
+        // so this function must not touch any global resources
+        // and callees must be thread-safe.
         if (!bm_display.IsOk()) {
             auto mapitem = imagetypes.find(image_type);
             if (mapitem == imagetypes.end()) return wxNullBitmap;
```
aardappel, line 49: below it accesses the global `sys`, though those functions do not appear to depend on `sys`. So maybe safer to make those function global so it is more obvious what they use.

aardappel:
Thanks!

PR #430, opened 2023-06-02
Avoid redundant code for refreshing Image trefc in imagelist

t2b3:
Also remove some empty lines

### DIFF
--- a/src/document.h
+++ b/src/document.h
@@ -198,8 +198,7 @@ struct Document {
             fos.Write(&vers, 1);
             sos.Write8(selected.xs);
             sos.Write8(selected.ys);
-            loopv(i, sys->imagelist) sys->imagelist[i]->trefc = 0;
-            rootgrid->ImageRefCount();
+            RefreshImageRefCount(true);
             int realindex = 0;
             loopv(i, sys->imagelist) {
                 Image &image = *sys->imagelist[i];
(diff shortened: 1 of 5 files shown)


src/document.h
```diff
@@ -724,6 +723,11 @@ struct Document {
         return GetBitmap().GetSubBitmap(r);
     }
 
+    void RefreshImageRefCount(bool includefolded = true) {
```
aardappel, line 726: Don't use default arguments unless on very frequently called functions or very exceptional arguments. Here, an explicit argument is better, it alerts the reader that a decision is being made.

src/grid.h
```diff
@@ -334,7 +334,7 @@ struct Grid {
     }
 
     Selection SelectAll() { return Selection(this, 0, 0, xs, ys); }
-    void ImageRefCount() { foreachcell(c) c->ImageRefCount(); }
+    void ImageRefCount(bool includefolded) { if(includefolded || !folded) foreachcell(c) c->ImageRefCount(includefolded); }
```
aardappel, line 337: good to spread more complex functions over multiple lines

src/document.h
t2b3: Ah thanks for the review and the good general advice.

src/grid.h
t2b3: Ok I did this - hopefully right

aardappel:
Thanks :)

PR #432, opened 2023-06-04
Avoid doing LoadImage twice for at least PNG files

t2b3:
The parser logic for the error case is now used for the normal parsing of the PNG file. This avoids to have LoadImage being called twice as it is later called when the buffer is converted to wxImage.

If LoadImage fails there, the replacement bitmap is now generated at the generic conversion step.

For JFIF, the parser logic is a little bit more complicated, so let LoadImage be called in this case.


aardappel:
Ahh, I guess previous the `Image` was used, now its unused, so this was still slowing down loading?

Now we still have slow JPEG loading though, but rather than writing a JPEG parser, I propose one further change (in a new PR): bump the file format version by 1, and prefix during writing each image by a length field. Then during reading, if a file of that version or higher, use the length field to skip any image, otherwise use the current method.

That will speed up JPEG loading also once users save the file in the new format

PR #433, opened 2023-06-04
Prefix image data with its length and increment TS_VERSION to 22

t2b3:
- This increments the TreeSheets file format version to 22.
- This adds a wxInt64/int64_t before the actual image data indicating its length.
- This updates the corresponding loading and saving procedures.
- This updates the file format specification.

### DIFF
--- a/TS/docs/file_format_spec.txt
+++ b/TS/docs/file_format_spec.txt
@@ -22,9 +22,9 @@ struct Image
 {
     char ident;   // marks start of image
     // Character value depends on image type, i.e. 'I' for PNG and 'J' for JPEG
+    double display_scale; // display scale of image
+    int64_t imagelen; // length of image (8 bytes)
     char image_data[len]; // whatever format wxImage::LoadFile uses
-    // to know len, you need to parse the png chunks
-    // sadly this means there's no way to read the Cells without doing so
 }
 
 struct Cell
--- a/src/document.h
+++ b/src/document.h
@@ -205,7 +205,9 @@ struct Document {
                 if (image.trefc) {
                     fos.PutC(image.image_type);
                     sos.WriteDouble(image.display_scale);
-                    fos.Write(image.image_data.data(), image.image_data.size());
+                    wxInt64 imagelen(image.image_data.size());
+                    sos.Write64(imagelen);
+                    fos.Write(image.image_data.data(), imagelen);
                     image.savedindex = realindex++;
                 }
             }
--- a/src/system.h
+++ b/src/system.h
@@ -302,37 +302,43 @@ struct System {
                         if (versionlastloaded < 9) dis.ReadString();
                         double sc = versionlastloaded >= 19 ? dis.ReadDouble() : 1.0;
                         vector<uint8_t> image_data;
-                        off_t beforeimage = fis.TellI();
-                        
-                        if(iti == 'I') {
-                            uchar header[8];
-                            fis.Read(header, 8);
-                            uchar expected[] = {0x89, 'P', 'N', 'G', '\r', '\n', 0x1A, '\n'};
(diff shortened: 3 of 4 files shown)


TS/docs/file_format_spec.txt
```diff
@@ -22,9 +22,9 @@ struct Image
 {
     char ident;   // marks start of image
     // Character value depends on image type, i.e. 'I' for PNG and 'J' for JPEG
+    double display_scale; // display scale of image
+    wxInt32 imagelen; // length of image (4 bytes)
```
aardappel, line 26: just say `int` like elsewhere in this file.. or `int64_t`.. best to keep this not wxwidgets specific where possible

src/document.h
```diff
@@ -205,7 +205,9 @@ struct Document {
                 if (image.trefc) {
                     fos.PutC(image.image_type);
                     sos.WriteDouble(image.display_scale);
-                    fos.Write(image.image_data.data(), image.image_data.size());
+                    wxInt32 imagelen(image.image_data.size());
```
aardappel, line 208: We don't expect images >2GB.. yet, to be correct, the result of `size()` (a 64-bit number) would need to be checked, to be correct.

Instead, since images are large, why don't we just store this length as 64-bit? No need for error checking, and future proof should anyone ever want to store larger images.

src/system.h
```diff
@@ -302,37 +302,44 @@ struct System {
...
-                        } else if(iti == 'J') {
-                            wxImage im;
-                            im.LoadFile(fis); // fixme: write own parser logic to avoid LoadFile
-                            if(!im.IsOk()) {
-                                return _(L"JPEG file is corrupted!");
+                        if(versionlastloaded >= 22) {
+                            wxInt32 imagelen = dis.Read32();
+                            image_data.resize(imagelen);
+                            fis.Read(image_data.data(), imagelen);
+                            if (!fis.IsOk()) image_data.clear();
```
aardappel, line 309: this line can be lifted after the if-then

t2b3: Ok I did this

src/document.h
t2b3: That might be a very huge image but you are absolutely right! Thanks :-)

TS/docs/file_format_spec.txt
t2b3: Ok, understood. This might for other parsers not being based on wxWidgets, of course.

src/system.h
```diff
@@ -302,37 +302,43 @@ struct System {
...
-                                if (memcmp(fourcc, "IEND", 4) == 0) break;
-                            }
-                        } else if(iti == 'J') {
-                            wxImage im;
-                            im.LoadFile(fis); // fixme: write own parser logic to avoid LoadFile
-                            if(!im.IsOk()) {
-                                return _(L"JPEG file is corrupted!");
+                        if(versionlastloaded >= 22) {
+                            wxInt64 imagelen = dis.Read64();
+                            image_data.resize(imagelen);
```
aardappel, line 307: This likely will need a `(size_t)`

src/system.h
```diff
@@ -302,37 +302,43 @@ struct System {
...
-                                dis.Read32();                   // skip CRC
-                                if (memcmp(fourcc, "IEND", 4) == 0) break;
-                            }
-                        } else if(iti == 'J') {
-                            wxImage im;
-                            im.LoadFile(fis); // fixme: write own parser logic to avoid LoadFile
-                            if(!im.IsOk()) {
-                                return _(L"JPEG file is corrupted!");
+                        if(versionlastloaded >= 22) {
+                            size_t imagelen = dis.Read64();
```
aardappel, line 306: please put a `(size_t)`, this will error if ever compiled on a 32-bit platform.

t2b3: Is it ok now?

aardappel:
Thanks!

PR #440, opened 2023-06-09
Implement Drag And Drop Copy out of TreeSheets

t2b3:
Also issue with GTK under Wayland with `wxDataObjectComposite` seems to be resolved.


aardappel:
What happens if you drag from treesheets to treesheets? I presume it still goes via the clipboard? would we want that to do the equivalent of copy+paste?

t2b3:
> Owner
Thanks for your reply! I still missed the part to place the cloned cell to `cellclipboard`. Now it works completly:

https://github.com/aardappel/treesheets/assets/3286756/7bbead7a-6730-4081-86df-15b3e5259d0e

t2b3:
https://github.com/aardappel/treesheets/assets/3286756/b891f9bf-dbea-44eb-be28-1bc54a941f9b

t2b3:
While Drag and Drop operates mainly on the stack, Clipboard operations operate mainly on the heap in WxWidgets. This is why I had to duplicate the logic.

aardappel:
Ok, LGTM.. nice new functionality!

jn64:
A side effect of this change is that Shift can't be used to extend selections with mouse any more (i.e. click on one cell, hold shift, click on another cell).

aardappel:
Hmm we should probably retain that functionality.. can we move the shift for dragging to a different modifier @t2b3 ?

t2b3:
> 

@aardappel and @jn64 
Could you explain to me where the difference of behaviour is between doing the selection with and without the Shift modifier pressed down? I could do the same also without the Shift modifier used or am I wrong? Thanks in advance.

t2b3:
I can see your point. I have to think about that :-)

t2b3:
> I tested briefly LibreOffice Calc 7.5 on Linux. They use Alt for drag-and-drop to other applications, and Shift-Alt for a move (cut) instead of copy. I also discovered that Alt and Ctrl are already used for dnd _within_ TS, which I wasn't aware of before. Alt-drag copies a cell, while Ctrl-drag moves a cell. Is it possible to have both internal and external dnd on Alt? If not, then I suggest external dnd should use two modifiers like Shift-Alt (and possibly Shift-Ctrl for move, mirroring the internal dnd behaviour).

See https://github.com/aardappel/treesheets/pull/447
I hope this is right (uses Shift-Alt).

PR #441, opened 2023-06-10
DRAFT: HTML Export: Inline images as base64

t2b3:


t2b3:
See here in action:
https://github.com/aardappel/treesheets/assets/3286756/652d8c1b-62d0-4afc-82e8-039c334b4551

aardappel:
First, is this HTML export run on any copy nowadays? I'd be concerned that this would get a lot slower with these base64 conversions in it, even though 99% of the time you are not using the HTML data.. wonder if thus an explicit "copy to clipboard with images" would be needed..

Second, for actual regular html export to file, this is going to export an image that is used 10x that many times in the file, in an inefficient format that makes the resulting file unreadable. Thus, for file html export it be much better to export images as png/jpeg files not inline.

t2b3:
> First, is this HTML export run on any copy nowadays? I'd be concerned that this would get a lot slower with these base64 conversions in it, even though 99% of the time you are not using the HTML data.. wonder if thus an explicit "copy to clipboard with images" would be needed..
> 
> Second, for actual regular html export to file, this is going to export an image that is used 10x that many times in the file, in an inefficient format that makes the resulting file unreadable. Thus, for file html export it be much better to export images as png/jpeg files not inline.

This needs more rework. Maybe I will find time to pick this up but for now I will not touch it directly.

PR #442, opened 2023-06-10
fix spelling: powerfull -> powerful

datgame:


aardappel:
Thanks :)

datgame:
I did this also to see how you'd diff 2 cts files :-) since i can't figure that out, besides exporting to one of the text formats first.

aardappel:
Yeah there's no way to do that.. probably html export would be best since that includes formatting changes. Would be nice to have a tool for that one day :)

PR #449, opened 2023-06-21
Link cells by their images

t2b3:


src/cell.h
```diff
@@ -427,13 +427,14 @@ struct Cell {
...
-                   bool forward) {
-        if (grid) best = grid->FindLink(s, link, best, lastthis, stylematch, forward);
+                   bool forward, bool image) {
+        if (grid) best = grid->FindLink(s, link, best, lastthis, stylematch, forward, image);
         if (link == this) {
             lastthis = true;
             return best;
         }
-        if (link->text.ToText(0, s, A_EXPTEXT) == text.t) {
+        if ((!image && link->text.ToText(0, s, A_EXPTEXT) == text.t) ||
```
aardappel, line 436: use `?:`

src/myframe.h
```diff
@@ -278,6 +278,7 @@ struct MyFrame : wxFrame {
             selmenu->AppendSeparator();
             MyAppend(selmenu, A_LINK, _(L"Go To &Matching Cell\tF6"));
             MyAppend(selmenu, A_LINKREV, _(L"Go To Matching Cell (Reverse)\tSHIFT+F6"));
+            MyAppend(selmenu, A_LINKIMG, _(L"Go To Cell with Matching Image\tSHIFT+F7"));
```
aardappel, line 281: that's a bit of a weird keyboard assignment. Why don't you make `L"Toggle Vertical Layout\tF7"` into `ALT+0`, then you can make the image search into `F7` (and have room to add a reverse in the future if anyone needs it)

t2b3: Thanks for the review and the helpful input! :+1: 
I revised the PR accordingly. I hope this fits better now.

AntonBogun:
Why exactly is this feature needed? Does it even need a keyboard shortcut?
This sounds like something that could be done with actual cell linking.

src/myframe.h
```diff
@@ -278,6 +278,8 @@ struct MyFrame : wxFrame {
             selmenu->AppendSeparator();
             MyAppend(selmenu, A_LINK, _(L"Go To &Matching Cell\tF6"));
             MyAppend(selmenu, A_LINKREV, _(L"Go To Matching Cell (Reverse)\tSHIFT+F6"));
+            MyAppend(selmenu, A_LINKIMG, _(L"Go To Cell with Matching Image\tF7"));
+            MyAppend(selmenu, A_LINKIMGREV, _(L"Go To Cell With Matching Cell (Reverse)\tSHIFT+F7"));
```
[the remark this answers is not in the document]
t2b3: Autsch! Thanks, a little bit of sloppiness... I revised it.

t2b3:
> Why exactly is this feature needed? Does it even need a keyboard shortcut? This sounds like something that could be done with actual cell linking.

The use case would be if you use the images / icons to organise your work, e.g. mark pending tasks with a certain icon or something to discuss with someone. This way you can cycle through all corresponding cells with the same icon/task.

AntonBogun:
Hmmm I suppose that makes sense. Though when I get to actually making a pull request for text-based linking I think I would change the image linking shortcuts to have ctrl+f7 and ctrl+shift+f7 respectively.

t2b3:
> > Why exactly is this feature needed? Does it even need a keyboard shortcut? This sounds like something that could be done with actual cell linking.
> 
> The use case would be if you use the images / icons to organise your work, e.g. mark pending tasks with a certain icon or something to discuss with someone. This way you can cycle through all corresponding cells with the same icon/task.



> 

For some reason Ctrl+... is handled the same as without Ctrl on Linux, for at least in combination with Function keys... :-) but I cannot explain. But sure there will be a way to work this out.

t2b3:
See the logs for the Windows CI...:
https://github.com/aardappel/treesheets/actions/runs/5337040536/jobs/9672549033?pr=449

AntonBogun:
Yeah it clearly says the issue is with lobster. If this was fixed in the lobster repository then it might be required to run the script or whatever to sync them up again.

t2b3:
> Yeah it clearly says the issue is with lobster. If this was fixed in the lobster repository then it might be required to run the script or whatever to sync them up again.

Could you do this? There is a Windows batch file sitting in the lobster folder to sync this stuff? Unfortunately I do not have Windows here.

AntonBogun:
Ok found it. Don't have time to mess around with this right now though, but it looks like it only does basic copy and makedir commands, so you shouldn't have much trouble following the commands from linux.

aardappel:
Ok, looks good, CI issue can be fixed elsewhere.

PR #451, opened 2023-06-24
DRAFT: Simplify CI generation of MacOS binaries

t2b3:
This makes CMake do most of the work by auto-generating the XCode files.


aardappel:
This is a nice simplification indeed..

Please capture the the commands in the CI also in local `.sh` files such that its easy for someone who wants to build locally to generate the xcode files.

Also, the way it was build before, the app file would have an icon, and include all the right files in the resource bundle.. have you tested this on a mac that its still doing that?

t2b3:
> This is a nice simplification indeed..
> 
> Please capture the the commands in the CI also in local `.sh` files such that its easy for someone who wants to build locally to generate the xcode files.

Ok, I split that part into `osx/generate_xcode.sh` and changed the workflow accordingly.

> 
> Also, the way it was build before, the app file would have an icon, and include all the right files in the resource bundle.. have you tested this on a mac that its still doing that?

No, all this stuff is pushed away. Is it not sufficient to have the binaries?

t2b3:
Ah all the resource stuff is missing... if I have time, I will look how to pack them together.

aardappel:
The CI is both for testing and making releases, so some people rely on it to grab a new app bundle.. that has to keep working.

The Xcode project was set up to create working app bundles, so that has to be tested to be true. Same with the icon. Also, some of those commands you deleted I believe were necessary to disable some kind of app store checking.

So basically, look at the bundle that the old CI used to produce. The CMake path has to produce the same bundle. If you don't have access to a mac I can test it on an M1, but you have to do the work making CMake produce the right stuff.

Generally, if you're changing stuff, assume things are there for a reason :)

PR #452, opened 2023-06-25
Allow non-privileged installation on Windows

t2b3:
This is useful for environments where users usually do not have elevated rights, like in business environments. This commit also changed the file association to the interactive user level.


aardappel:
Nice!

PR #455, opened 2023-06-30
Allow to insert tag in multiple cells

t2b3:


src/document.h
```diff
@@ -2253,11 +2253,13 @@ struct Document {
         int i = 0;
         for (auto tagit = tags.begin(); tagit != tags.end(); ++tagit)
             if (i++ == tagno) {
-                Cell *c = selected.GetCell();
-                if (!c) return OneCell();
-                c->AddUndo(this);
-                c->text.Clear(this, selected);
-                c->text.Insert(this, tagit->first, selected);
+                selected.g->cell->AddUndo(this);
+                loopallcellssel(c, true) {
```
aardappel, line 2257: when is this useful? I can't imagine wanting to fill a whole grid all with the same tag?

t2b3:
Actually this can be the use case if you want to fill a grid or some cells with default tags and then to modify only some of them later (e.g. replace them with another tag).

aardappel:
Hmm, I guess it doesn't hurt to add?

PR #458, opened 2023-06-30
Copy only selection if cell is in TextEdit mode

t2b3:


aardappel:
Can you explain what the effect is of NOT setting `sys->clipboardcopy` ?

t2b3:
When the string from the system clipboard and TreeSheets internal clipboard do not match, TreeSheets will not copy the cell from the internal cellclipboard. When we copy only a part of the cell text and paste it afterwards in TreeSheets, we only want that part and not the whole cell from the cellclipboard.

t2b3:
@aardappel 
Is this ok for you?

t2b3:
Behaviour without this patch (complete cell is copied):
https://github.com/aardappel/treesheets/assets/3286756/29dc1ec5-48d7-4adf-8fac-b5289e0829d6

Behaviour with this patch (selected text is copied):
https://github.com/aardappel/treesheets/assets/3286756/1b356f01-3b5a-4c07-b847-605011dcfac7

aardappel:
Ok, thanks!

PR #464, opened 2023-07-08
Add clang-tidy option, default OFF

nunotexbsd:
Add a debug option for clang linter.
Default OFF as not all systems install this extra tool.


aardappel:
Thanks!

PR #467, opened 2023-07-11
Provide similiar experience when Return key is pressed

t2b3:
This provides a similiar experience when Return key is pressed in the search and the replace text control.

This implements it for MSW and the other ports.

Also there is now a keybinding to jump directly into the replace text control.


t2b3:
Demonstration:


https://github.com/aardappel/treesheets/assets/3286756/7df6a846-a035-4185-855a-cefe65509832

src/myframe.h
```diff
@@ -446,7 +446,8 @@ struct MyFrame : wxFrame {
         semenu->AppendCheckItem(A_CASESENSITIVESEARCH, _(L"Case-sensitive search"));
         semenu->Check(A_CASESENSITIVESEARCH, sys->casesensitivesearch);
         MyAppend(semenu, A_SEARCHNEXT, _(L"&Go To Next Search Result\tF3"));
-        MyAppend(semenu, A_REPLACEONCE, _(L"&Replace in Current Selection\tCTRL+k"));
+        MyAppend(semenu, A_REPLACEF, _(L"&Replace\tCTRL+r"));
```
aardappel, line 449: Was CTRL+r available? CTRL+h is more common for replace if its available

t2b3: Ctrl+h is mapped to backspace on Linux/wxGTK.

t2b3: It is not mapped to any action (but Ctrl+Shift+r is mapped to Reset text styles)

aardappel: ctrl+h is backspace maybe in the shell, but not in modern editors like vscode: https://code.visualstudio.com/docs/getstarted/keybindings

We should probably follow that example

t2b3: Unfortunately it seems to be the case also on UI. I will test it with Ctrl+h and I hope for some kind of superposition...

t2b3: Tested it. But Ctrl+h is bound to backspace in wxGTK per system default and cannot be overwritten by the accelerator keys of the menu item.

aardappel: It seem possible: https://www.devxperiences.com/pzwp1/2023/06/10/wxwidgets-on-linux/
Some other posts hint at it being a GTK configuration issue.

Either way, pretty important that it is CTRL+h on Windows, and most Linux apps use it too. So I'd say to just universally set it to CTRL+h, and given that this is a new binding, either Linux users won't miss it, or they can change their GTK settings.

t2b3: Ok I found a way that prevents the translation from Ctrl+letter into ASCII control characters (see wxKeyEvent documentation). It works now with Ctrl+h.

aardappel:
LGTM!

PR #469, opened 2023-07-15
Remember and load TreeSheets file with zoom level

t2b3:
Rationale: In a TreeSheets document with many levels (nested grids etc.), it is an improvement for the convenience to already load the document with the zoom level that was set when the document has been saved.

### DIFF
--- a/src/document.h
+++ b/src/document.h
@@ -20,6 +20,7 @@ struct Document {
     int originx, originy, maxx, maxy, centerx, centery;
     int layoutxs, layoutys, hierarchysize, fgutter;
     int lasttextsize, laststylebits;
+    int initialzoomlevel;
     Cell *curdrawroot;  // for use during Render() calls
     Vector<UndoItem *> undolist, redolist;
     Vector<Selection> drawpath;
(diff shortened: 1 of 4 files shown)


src/document.h
```diff
@@ -663,6 +666,7 @@ struct Document {
         if (hover.g) hover.g->DrawHover(this, dc, hover);
         if (scaledviewingmode) { dc.SetUserScale(1, 1); }
         if (scrolltoselection) {
+            Zoom(initialzoomlevel, dc);
```
aardappel, line 669: this would appear to do it on every draw, not just the first?

aardappel:
Hmm.. this is useful for a person that frequently opens the same document, but not so great if you share documents with others.. I guess the first usecase is more common?

src/document.h
t2b3: `scrolltoselection` is switched to `false` after the first draw.

t2b3:
> 

I think so, too.

t2b3:
> Hmm.. this is useful for a person that frequently opens the same document, but not so great if you share documents with others.. I guess the first usecase is more common?

I think so, too.

PR #470, opened 2023-07-16
Do not zoom to initialzoomlevel when DPI changes

t2b3:
Do not zoom with initial zoom level in other occurences when `scrolltoselection` is invoked, e.g. `OnDPIChanged`
→ Separate `scrolltoselection` signal from newly added `initialzoomtolevel` signal

### DIFF
--- a/src/document.h
+++ b/src/document.h
@@ -665,8 +665,11 @@ struct Document {
         DrawSelect(dc, selected);
         if (hover.g) hover.g->DrawHover(this, dc, hover);
         if (scaledviewingmode) { dc.SetUserScale(1, 1); }
-        if (scrolltoselection) {
+        if (initialzoomlevel) {
             Zoom(initialzoomlevel, dc);
+            initialzoomlevel = 0;
+        }
+        if (scrolltoselection) {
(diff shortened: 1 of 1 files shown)


src/document.h
```diff
@@ -665,8 +667,11 @@ struct Document {
         DrawSelect(dc, selected);
         if (hover.g) hover.g->DrawHover(this, dc, hover);
         if (scaledviewingmode) { dc.SetUserScale(1, 1); }
-        if (scrolltoselection) {
+        if (initialzoomtoselection) {
```
aardappel, line 670: Why don't you just do this if `initialzoomlevel` is not 0 or -1 or whatever? Adding extra bools for each little thing gets very confusing, you want as little of them as possible :)

t2b3: Thanks for your valuable feedback! I revised the commit accordingly and I hope that it fits better now.

t2b3: > Why don't you just do this if `initialzoomlevel` is not 0 or -1 or whatever? Adding extra bools for each little thing gets very confusing, you want as little of them as possible :)

This a lot smarter! Thanks :-)

PR #474, opened 2023-07-19
Auto-zoom to search match

t2b3:


t2b3:
> Does this always zoom? I mean, if a search result is already visible, but not zoomed in, it should probably not zoom in? The idea behind ScrollOrZoom is to make something visible if it wasn't, but it should do the least amount of scrolling or zooming to achieve that.

I see the rationale behind it and I agree to leave it as-is. Thanks for your reply!

PR #481, opened 2023-08-01
Slightly modernise HTML export

jn64:
The current doctype triggers [quirks mode](https://developer.mozilla.org/en-US/docs/Web/HTML/Quirks_Mode_and_Standards_Mode), which is for _pre_-IE6 compatibility.

I tested all 3 HTML export (table, bullet, outline) of tutorial.cts to make sure they look the same in standards mode (in Firefox 115 on Linux).

The table style looks 100% the same, pixel-perfect. The outline style has negligible alignment differences (some text looks higher by 1px).

The bullet style has 1 visible difference: in quirks mode, nested lists (`li > ul`) with no text in the parent li has extra padding on top, which I would say is a bug/unintended (from TS' point of view), since there is no extra padding anywhere else.

Screenshot comparing nested lists in quirks mode and standards mode

The `font-weight: normal` is added to reset the font weight in nested tables, because standards mode properly inherits font styles in tables which quirks mode did not.

Benefit of this change is users who edit the HTML export (e.g. to apply additional styling) should have more predictable modern behaviour.

### DIFF
--- a/src/document.h
+++ b/src/document.h
@@ -852,18 +852,17 @@ struct Document {
                 case A_EXPHTMLB:
                 case A_EXPHTMLO:
                     dos.WriteString(
-                        L"<!DOCTYPE HTML PUBLIC \"-//W3C//DTD HTML 3.2 Final//EN\">\n"
+                        L"<!DOCTYPE html>\n"
                         L"<html>\n<head>\n<style>\n"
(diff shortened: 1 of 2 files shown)


aardappel:
Thanks for modernizing :)

PR #485, opened 2023-08-08
Move newly added functions to `mywxtools.h`

t2b3:
Similiar functions are stored in `mywxtools.h`, so better to keep it that way.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -275,15 +275,6 @@ enum {
     TEXT_SEP = 2,
     TEXT_CHAR = 1
 };
-inline bool IsWordSep(wxChar ch) {
-    //represents: !"#$%&'()*+,-./    :;<=>?@    [\]^    {|}~    `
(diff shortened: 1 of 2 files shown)


src/mywxtools.h
```diff
@@ -179,3 +179,14 @@ static wxBitmap ConvertBufferToWxBitmap(vector<uint8_t> &buf, wxBitmapType bmt)
     wxBitmap bm(im, 32);
     return bm;
 }
+
+inline static bool IsWordSep(wxChar ch) {
```
aardappel, line 183: Are these used in multiple files? If they are only used in the selection code, it be best to move them close to where they are used

t2b3: Hi :-) Thanks for your feedback. It looks like the callers are only in `selection.h` itself, so I updated the PR/commit.

t2b3:
@aardappel Thanks for the merge. Have a nice day!

AntonBogun:
ahh I didn't mention but yes those functions will end up being used again in text.h since I'm planning to fix how selection of a word with a mouse happens...

PR #487, opened 2023-08-09
gitignore script folder

AntonBogun:
Add TS/scripts to gitignore since if there are any user scripts in there they will appear on git status as untracked.
Any new example scripts can be added directly, and a nested git repository can be setup within the scripts folder to have source control over user scripts.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,4 @@ TreeSheets/x64
 *.ipdb
 *.iobj
 TreeSheets/gettext_bin
+TS/scripts


aardappel:
There are files in there that are tracked by git, so this is not a great idea.

Not sure what the right solution is, maybe a `scripts/user` folder?

AntonBogun:
Existing files in the repository do not stop being tracked, in fact I think it even allows adding them if they are changed.
The problem with doing the user folder is that i don't think treesheets will add them as scripts that can be used. I suppose it could be changed in code to support this, but I don't see why simply ignoring the whole scripts folder is a bad idea.

aardappel:
Yeah, still not great as you won't see new files.

This is only an issue for people who are actively using TreeSheets from a git development dir, so I don't think is a good tradeof.

aardappel:
Closing for now, until new opinions / new changes come in..

PR #489, opened 2023-08-09
Selection history

AntonBogun:


AntonBogun:
I've been using it successfully for over a month now and I think ironed out most bugs, though I wouldn't be surprised if an edge case is still there somewhere.
Made Mouse4/Mouse5 act as back/forth in the selection history, to match how web browsers work (and selection history in IDEs)

AntonBogun:
A thing to consider is that the current way the selection jumps around and leaves the target cell on the edge of the screen, is maybe not the best, though I'm not entirely sure how (and if) it should be changed.

AntonBogun:
The functionality is being able to go back to previous cells when performing a search. If only the last selection is kept track of, then returning to the old selection would only work in case of going to only one search result, or by not counting the selections during the search. But even then, if one decides to look around the search result and selects something, the old selection would be lost.
This feature makes it much more comfortable to work in a subset of the treesheet by making transitions significantly faster.

As for fixing the selection, I could exchange the path checking code with an existing function if it is needed.

AntonBogun:
The redo is needed because without it going back in history will be irreversible and thus the hassle in using it will probably be high enough that it would be easier to manually go back (since if you overshot you couldn't go back to latest selection etc). At that point, the feature will probably be niche enough that there's no point implementing it.

I don't think the PR can be simplified much beyond simple code shortening, all the logic (besides the already implemented elsewhere selection path walking) has to be there as far as I'm aware.

aardappel:
If it can't be simplified then I am not sure the added complexity is worth the functionality..

aardappel:
Closing for now, until new opinions / new changes come in..

PR #494, opened 2023-08-24
Reorganise selection in `InitWith` code

t2b3:
- Make it more compact.
- When cell found, jump to end of routine instead of looping through the rest of the cells in the grid.

### DIFF
--- a/src/document.h
+++ b/src/document.h
@@ -139,19 +139,17 @@ struct Document {
 
     void InitWith(Cell *r, wxString filename, Cell *ics, int xs, int ys) {
         rootgrid = r;
-        if (ics) {
-            Grid* ipg = ics->parent->grid;
-            if (ipg) {
-                foreachcellingrid(c, ipg) {
-                    if (c == ics) {
-                        SetSelect(Selection(ipg, x, y, xs, ys));
-                    }
-                }
+        Grid *ipg = ics ? ics->parent->grid : nullptr;
+        if (ipg) {
+            foreachcellingrid(c, ipg) if (c == ics) {
+                SetSelect(Selection(ipg, x, y, xs, ys));
+                goto cellisselected;
             }
         } else {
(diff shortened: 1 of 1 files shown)


aardappel:
This changes the logic of whats happening, before, if ipg was null, it wouldn't call SetSelect, now it does, is this intentional?

The new code is harder to follow, so unless breaking the loop is important, it is probably not an improvement.

t2b3:
> This changes the logic of whats happening, before, if ipg was null, it wouldn't call SetSelect, now it does, is this intentional?
> 
> 
> 
> The new code is harder to follow, so unless breaking the loop is important, it is probably not an improvement.
> 
> 

Hi Wouter!
The logic is not inverted. If ipg is nullptr, SetSelected will not be called. And I consider that breaking the loop is an improvement. Can we agree to leave most of it but add the possibility of breaking the loop?
Kind regards

aardappel:
If ipg is null, it calls the SetSelect in the else branch?

t2b3:
> If ipg is null, it calls the SetSelect in the else branch?

Ok this was wrongly expressed. In the else branch there is `SetSelect`, for only for the first cell in the top-left.

t2b3:
On the other side, the case of `ics` evaluating to true and `ipg` evaluating to false (nullptr) is now considered in this PR.

t2b3:
Ok now I got what you meant. Would this be a problem to call `SetSelect` to select the first cell on the top left if `ipg` was `nullptr`?

t2b3:
Ok I agree better  to keep it the old way.

t2b3:
> This changes the logic of whats happening, before, if ipg was null, it wouldn't call SetSelect, now it does, is this intentional?
> 
> The new code is harder to follow, so unless breaking the loop is important, it is probably not an improvement.

Short answer: No, it was not really intentional :-)

t2b3:
Sorry for the noise!

PR #496, opened 2023-08-27
Update scrollbar after zoom on wxGTK

t2b3:


src/document.h
```diff
@@ -745,6 +745,15 @@ struct Document {
         #ifndef __WXMSW__
         if (sw) sw->Refresh(false);
         #endif
+        #ifdef __WXGTK__
+        if (sw) {
+            // wxWidgets (wxGTK) does not always automatically update the scrollbar
```
aardappel, line 750: This calls both the above `Refresh` and your new code GTK, are you sure both are needed?

t2b3: Yes, unfortunately. I tested it with a TreeSheets with nested grids and this is the way I found out to remove the bug that the scrollbar is at default position (0,0) when zoomed in and out.

t2b3: Bug demonstrated:

https://github.com/aardappel/treesheets/assets/3286756/cb34f04c-f553-40b1-8fd0-cc127abaa8fa

t2b3: With  this PR

https://github.com/aardappel/treesheets/assets/3286756/2bbd8496-8ea6-4595-afff-07a6a60eebbe

t2b3: @aardappel Do you agree with the rationale behind this PR?

aardappel: Yes, thanks!

PR #497, opened 2023-08-27
Align selection to center

t2b3:
This aligns the selection to center when selection was out of view before.

Trade-off:
🇨🇭 better placement of cells when zooming in and out
➖ bigger "jumps" when scrolling to the next cell that is outside of the view(port) as the selection will be placed to center

### DIFF
--- a/src/document.h
+++ b/src/document.h
@@ -280,10 +280,10 @@ struct Document {
                     sw->SetScrollbars(1, 1, layoutxs, layoutys,
                                       r.width > canvasw || r.x < originx
                                           ? r.x
-                                          : r.x + r.width > maxx ? r.x + r.width - canvasw : curx,
+                                          : r.x + r.width > maxx ? r.x + r.width / 2 - canvasw / 2 : curx,
                                       r.height > canvash || r.y < originy
                                           ? r.y
-                                          : r.y + r.height > maxy ? r.y + r.height - canvash : cury,
+                                          : r.y + r.height > maxy ? r.y + r.height / 2 - canvash / 2 : cury,
                                       true);
                     RefreshReset();
                     return true;


aardappel:
I think adjustments for out of view should always move as little as possible. You want users to be able to keep context.

t2b3:
Before:

https://github.com/aardappel/treesheets/assets/3286756/f23aa47c-0662-43cb-b649-c3cef6424b66

After:

https://github.com/aardappel/treesheets/assets/3286756/2096dab6-c2d1-4225-abc8-7522b4b1670d

You can see that before, when zooming out, the selection is always at the edge (`r.x + r.width - canvasw`).

t2b3:
One can also argue that with this PR, one can see more context below.

https://github.com/aardappel/treesheets/assets/3286756/91e751fe-8c25-4635-8948-8702c2ac1667

t2b3:
On the other side, yeah it could be annoying if it jumps to much....

AntonBogun:
The solution I came up with on my own fork is to blink (invert) the edges of the selected cell twice on a "discontinuous" transition, such as for example going to a search result.
I could try to make that into a PR if there's a want.

aardappel:
@t2b3 
I don't think it matters too much for the zooming out case (first example), since you had no context anyway.

For the second example (extending selection) it seems useful, but it is also a bit arbitrary that it jumps by random amounts based on the current size.. if you wanted to show more context, it be better to do the old behavior `- one_grid_cell_height` (for this specific case).

Where it be much worse if you jump between e.g. search results.. current result is on middle of page, next result the existing code scrolls the minimum amount possible, giving a) the least amount of scrolling, and b) the most amount of change the old new location are both on screen. Your new code will simply put the new result in the middle of the screen, which would be quite bad.

So, if you wanted to make these improvements, I'd identify callers of this function, and maybe pass a flag for different possible optimal behaviors based on the operations that triggers them. Again, in case of doubt, least amount of scrolling is best.

@AntonBogun a blink to notify the user of a new selection generally sounds as nice feedback, if this can be done without complicating the code a lot (in terms of having to do this anim over time). Does it delay when the user can interact until the anim is done?

AntonBogun:
@aardappel It works similarly to how the cursor blink works, and the rendering code lives in the selection drawing function. The idea is to activate a timer that will run 4 times which will re-render the selection with the outline on/off.

This is a video of it in action, in the beginning I am using my Selection History feature to navigate which I've made to trigger the outline.

https://github.com/aardappel/treesheets/assets/70374466/156602ee-693e-4069-acdf-71d9557be7f7

I've been using this for a couple weeks now and it works really well thus far.

t2b3:
Thanks for your reply.

> @t2b3 
> 
> I don't think it matters too much for the zooming out case (first example), since you had no context anyway.
> 

It might matter more when you zoom out of a cell in the center of a grid, for example?

> 
> For the second example (extending selection) it seems useful, but it is also a bit arbitrary that it jumps by random amounts based on the current size.. if you wanted to show more context, it be better to do the old behavior `- one_grid_cell_height` (for this specific case).
> 
> 
> 
> Where it be much worse if you jump between e.g. search results.. current result is on middle of page, next result the existing code scrolls the minimum amount possible, giving a) the least amount of scrolling, and b) the most amount of change the old new location are both on screen. Your new code will simply put the new result in the middle of the screen, which would be quite bad.
> 
> 

- Do I get it right that you implicitly assume that the context is on the left and on the top of the selection? This might not be always the case when you have grids with cells around the selections when they are filled with content?


> 
> So, if you wanted to make these improvements, I'd identify callers of this function, and maybe pass a flag for different possible optimal behaviors based on the operations that triggers them. Again, in case of doubt, least amount of scrolling is best.
> 
> 
> 
> @AntonBogun a blink to notify the user of a new selection generally sounds as nice feedback, if this can be done without complicating the code a lot (in terms of having to do this anim over time). Does it delay when the user can interact until the anim is done?

AntonBogun:
I think current logic just puts the whole cell text into view, regardless of what else is in the cell. This makes complete sense since a cell may be larger than even the screen.

I think any other logic will have it's edge cases when it doesn't do what would be ideal, but I suspect the current logic has as little of them as possible. The only problem is that the user doesn't really know where the selection went to if it transfered to a completely different part of the treesheet, but doing a blink remedies that.

t2b3:
I agree that it might be a good idea to differentiate between different situations and to control it with a flag.

AntonBogun:
This sounds too complicated to me, thus far the existing logic but with blink worked just fine.

aardappel:
@AntonBogun while the idea of blinking is good, the really fat border blink you have there would be too invasive for me.. I was expecting more like blinking the actual selection box only.

@t2b3 
> Do I get it right that you implicitly assume that the context is on the left and on the top of the selection? This might not be always the case when you have grids with cells around the selections when they are filled with content?

No, a similar situation could occur with different configurations. I simply want the minimum amount of movement from the current view in all cases. That still means the current code can be improved, but it really depends on the case.

t2b3:
> @AntonBogun while the idea of blinking is good, the really fat border blink you have there would be too invasive for me.. I was expecting more like blinking the actual selection box only.
> 
> @t2b3
> 
> > Do I get it right that you implicitly assume that the context is on the left and on the top of the selection? This might not be always the case when you have grids with cells around the selections when they are filled with content?
> 
> No, a similar situation could occur with different configurations. I simply want the minimum amount of movement from the current view in all cases. That still means the current code can be improved, but it really depends on the case.

Ok, thanks for your reply. I am not so sure any more whether the added complexity is it worth...

t2b3:
@aardappel @AntonBogun 
After giving it some thought, I decided to close this PR considering all the points you mentioned.
Thanks for having taken your time and for sharing the ideas! Given all the alternatives, the current implementation may the best under these constraints.

PR #498, opened 2023-08-31
Copy solely the text string when TextEdit is active

t2b3:
When the cell is in TextEdit mode and thus the cell text partly or wholly selected, only the selected text string shall be copied to clipboard.

This is particularly useful when the user only wants to copy a part of the cell text without having the HTML formatting.


t2b3:
@aardappel 
Thanks for merging! 👍
Have a nice day! 😀

PR #499, opened 2023-09-01
Copy solely the text string when TextEdit is active (DnD).

t2b3:
Same as 7f68776a9e072520c735479929efecd0d58f362d but for Drag and Drop operation.

### DIFF
--- a/src/document.h
+++ b/src/document.h
@@ -410,6 +410,19 @@ struct Document {
         Refresh();
     }
 
+    auto CopyEntireCells(wxString &s) {
+        sys->clipboardcopy = s;
+        wxString html = selected.g->ConvertToText(selected, 0, A_EXPHTMLT, this);
+        auto htmlobj = 
+        #ifdef __WXGTK__
+            new wxCustomDataObject(wxDF_HTML);
+            htmlobj->SetData(html.Len(), html);
+        #else
+            new wxHTMLDataObject(html);
+        #endif
+        return htmlobj;
+    }
+
     void Copy(int mode) {
         Cell *c = selected.GetCell();
         sys->clipboardcopy = wxEmptyString;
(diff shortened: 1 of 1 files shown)


t2b3:
> That seems very similar code then.. should that be pulled into a function?

This is a fantastic idea. I have merged the code for clipboard and drag and drop in into one main control flow.
What do you think?

aardappel:
That required so many conditionals, that is definitely worse :)

What I meant is that I saw this code being identical in 2 places:

```
                        sys->clipboardcopy = s;
                        wxString html = selected.g->ConvertToText(selected, 0, A_EXPHTMLT, this);
                        auto *htmlobj = 
                        #ifdef __WXGTK__
                            new wxCustomDataObject(wxDF_HTML);
                        htmlobj->SetData(html.Len(), html);
                        #else
                            new wxHTMLDataObject(html);
                        #endif
```

So that could possibly be in a function, and maybe other parts of the code too.

t2b3:
> That required so many conditionals, that is definitely worse :)
> 
> What I meant is that I saw this code being identical in 2 places:
> 
> ```
>                         sys->clipboardcopy = s;
>                         wxString html = selected.g->ConvertToText(selected, 0, A_EXPHTMLT, this);
>                         auto *htmlobj = 
>                         #ifdef __WXGTK__
>                             new wxCustomDataObject(wxDF_HTML);
>                         htmlobj->SetData(html.Len(), html);
>                         #else
>                             new wxHTMLDataObject(html);
>                         #endif
> ```
> 
> So that could possibly be in a function, and maybe other parts of the code too.

@aardappel 
I agree. What about this iteration then? :-)

aardappel:
I am not sure what you're asking.

t2b3:
> I am not sure what you're asking.

Ok. I updated the PR (see commit). Is it ok now?

src/document.h
```diff
@@ -496,6 +480,23 @@ struct Document {
         return;
     }
 
+    #ifdef __WXGTK__
```
aardappel, line 483: why did you split it up by duplicating the code? the old code was better

t2b3: The return types differ between __WXGTK__ and other platforms.

t2b3: Ok I changed that. But I need to set the return type and differentiate between _WXGTK_ and others as also within the function body, there is a second differentiation then.

t2b3:
@aardappel 
Is it ok now?

src/document.h
```diff
@@ -496,6 +480,22 @@ struct Document {
         return;
     }
 
+#ifdef __WXGTK__
```
aardappel, line 483: just use a `-> auto` return type

src/document.h
```diff
@@ -496,6 +480,22 @@ struct Document {
...
 
+#ifdef __WXGTK__
+    wxCustomDataObject *CopyEntireCells(wxString &s) {
+#else
+    wxHTMLDataObject *CopyEntireCells(wxString &s) {
+#endif
+        sys->clipboardcopy = s;
+        wxString html = selected.g->ConvertToText(selected, 0, A_EXPHTMLT, this);
+        #ifdef __WXGTK__
+        wxCustomDataObject *htmlobj = new wxCustomDataObject(wxDF_HTML);
```
aardappel, line 491: `auto *htmlobj =` outside the ifdef like the original..  and indent

t2b3: Thanks for your reply. I also left the asterik now out because it is also deduced that it is a pointer type.

src/document.h
t2b3: And put the helper function above that did the trick!
Thanks for your feedback! :-)

t2b3:
@aardappel 
Does the revision fulfill your requirements?

aardappel:
Much nicer, thanks!

t2b3:
> Much nicer, thanks!

Thank you for your review making it much better!

PR #503, opened 2023-09-08
Emit cellcolor and/or textcolor in X(HT)ML

t2b3:


t2b3:
@aardappel This is a small fix I discovered while fixing the other bug.

src/cell.h
```diff
@@ -260,9 +260,9 @@ struct Cell {
             if (text.stylebits & STYLE_BOLD) style += L"font-weight: bold;";
```
aardappel, line 260: could you make a similar change for uses of `stylebits` ?

t2b3: Yes, I can try.

t2b3:
@aardappel 
What do you think of these two commits?
Caution: This will work fine with a HTML export of the whole TreeSheets document, but not with a partial one like when you copy the HTML export of the selection to clipboard (because the inheritered information is outside of the selection).

src/document.h
```diff
@@ -879,8 +879,7 @@ struct Document {
                         L"<html>\n<head>\n<style>\n"
                         L"body { font-family: sans-serif; }\n"
                         L"table, th, td { border: 1px solid #A0A0A0; border-collapse: collapse;"
-                        L" padding: 3px; vertical-align: top; font-weight: normal;"
-                        L" font-style: normal; font-family: sans-serif; }\n"
```
aardappel, line 883: why was this deleted?

src/cell.h
```diff
@@ -237,12 +237,12 @@ struct Cell {
                 str.Prepend(wxString() << text.stylebits);
                 str.Prepend(L" stylebits=\"");
```
aardappel, line 238: please also check for these `stylebits`

src/cell.h
```diff
@@ -257,12 +257,15 @@ struct Cell {
             str.Append(L"</cell>\n");
         } else if (format == A_EXPHTMLT) {
             wxString style;
-            if (text.stylebits & STYLE_BOLD) style += L"font-weight: bold;";
-            if (text.stylebits & STYLE_ITALIC) style += L"font-style: italic;";
-            if (text.stylebits & STYLE_FIXED) style += L"font-family: monospace;";
-            if (cellcolor != doc->Background())
+            if (parent ? (text.stylebits & STYLE_BOLD) != (parent->text.stylebits & STYLE_BOLD) : true)
```
aardappel, line 260: can be simplified to `!parent || (text.stylebits & STYLE_BOLD) != (parent->text.stylebits & STYLE_BOLD)`

t2b3: Ah thanks! (y)

src/document.h
t2b3: Because if we do inheritance and we have parent and child share e.g. same font-style that is not the default one (e.g. bold instead of normal), then the code will not add that font-style again for the child as it shall be inherited (this is the goal). But because no font-style is supplied in the child's style attribute then but a default one defined in the style section in the header, that one is prioritized over the inherited one. In order to avoid that situation, do not create this overriding styles in the header.

src/cell.h
t2b3: Ok please see last commit. It XORs the stylebits from child and parent. Only if they differ, stylebits will be emitted in XML export.

t2b3:
@aardappel So I added two commits. They are good to go for a next review round from my side.
I am looking forward for your feedback.

aardappel:
LGTM!

t2b3:
> LGTM!

Thanks for your highly valuable review and your time spent! I highly appreciate this. I wish you a wunderful weekend!

PR #504, opened 2023-09-08
Do not emit empty style attribute

t2b3:


t2b3:
@aardappel This should save some empty attributes when style value is empty. This is ready to go from my side. Could you review and merge if it is ok? Thanks in advance.

PR #513, opened 2023-09-19
De-duplicate code (Home/End navigation)

t2b3:


src/document.h
```diff
@@ -1990,39 +1990,20 @@ struct Document {
                 ZoomOutIfNoGrid(dc);
                 return nullptr;
 
-            // FIXME: this functionality is really SCHOME, SHOME should be within line
-            case A_SHOME:
+            case A_SHOME: case A_SEND: case A_CHOME: case A_CEND: case A_HOME: case A_END: {
```
t2b3, line 1993: Is it admissible? I say this is not the code style, but it saves some lines...

t2b3:
@aardappel 
Do you think the extra abstraction is justified in this case? Thanks in advance for looking into this.

aardappel:
Yes, I guess this one is fine :)

t2b3:
> Yes, I guess this one is fine :)

Thanks for having looked into this! Have a nice evening :-)
Best regards
Tobias

PR #514, opened 2023-09-19
Merge functions (Refresh/RefreshReset/RefreshHover)

t2b3:
This commit unifies `RefreshReset`, `RefreshHover` and `Refresh`.

Reasons:
- `RefreshReset` only called `Refresh`.
- `RefreshHover` was only called by `Refresh`.


t2b3:
@aardappel Hi
This one is not quite like the previous PR you have commented before. Could you look into this one? Thanks in advance.

PR #515, opened 2023-09-19
Pack declaration of multiple variables

t2b3:
This commit packs variable declaration of the same type into one line.

### DIFF
--- a/src/document.h
+++ b/src/document.h
@@ -1331,8 +1331,7 @@ struct Document {
             case A_FILTEROFF: SetSearchFilter(false); return nullptr;
 
             case A_CUSTKEY: {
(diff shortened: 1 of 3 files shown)


aardappel:
This one is ok, but a lot of people nowadays prefer a single variable per line.
Especially for member variables, it is good to have a single variable per line with an `=` initialization, rather than have the initialization in the constructor, where possible.

t2b3:
> This one is ok, but a lot of people nowadays prefer a single variable per line. Especially for member variables, it is good to have a single variable per line with an `=` initialization, rather than have the initialization in the constructor, where possible.

Thanks for your feedback. This might be another action point to change :smile:

PR #516, opened 2023-09-19
De-duplicate code (Filter)

t2b3:
This commit re-structures the filter logic in order to avoid redundant lines.

### DIFF
--- a/src/document.h
+++ b/src/document.h
@@ -1304,29 +1304,22 @@ struct Document {
             }
 
             case A_FILTER5:
-                editfilter = 5;
-                ApplyEditFilter();
-                return nullptr;
             case A_FILTER10:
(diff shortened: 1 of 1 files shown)


aardappel:
Same here.. you get 6 `case` immediately followed by 6 more.
If you wanted to improve this code, I'd add the delta value as an argument to this function so it can set it for you, and have it return the `nullptr`.

t2b3:
> Same here.. you get 6 `case` immediately followed by 6 more.
> 
> If you wanted to improve this code, I'd add the delta value as an argument to this function so it can set it for you, and have it return the `nullptr`.

Hello Wouter
Thanks for your review. I will think about your proposal, as we also have the editfilter incremented and decremented.

Best regards
Tobias

PR #517, opened 2023-09-19
De-duplicate code (Color dropdowns)

t2b3:
This commit avoids redundant lines for calling the color dropdown popup.

### DIFF
--- a/src/document.h
+++ b/src/document.h
@@ -1220,14 +1220,17 @@ struct Document {
                 return nullptr;
 
             case A_OPENCELLCOLOR:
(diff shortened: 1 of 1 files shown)


aardappel:
I'd say the new code is actually longer and more complicated to follow.. I am generally a fan of reducing duplicated code, but one-liners like this don't hurt much. 

You always have the weigh the cost of the duplication against the cost of the abstraction (extra functions, variables, switches..). Another concern is how far away from eachother are the duplications? If far, its easier to make errors when modifying them, but in this case they're right next to eachother too.

If you're interested in reducing duplication, try this tool, which takes these factors into account: https://strlen.com/sloppy/

PR #518, opened 2023-09-19
De-duplicate code (Text controls)

t2b3:
The code introduced by this commit avoids redundant lines.


aardappel:
See my explanation on the other PR on why this might not be a good de-duplication.

PR #522, opened 2023-09-26
[PLEASE DO NOT MERGE YET] Workaround is no longer needed for wxWidgets 3.3.0

t2b3:
Please do not merge yet as wxWidgets 3.3.0 is not yet released.

### DIFF
--- a/src/document.h
+++ b/src/document.h
@@ -413,13 +413,7 @@ struct Document {
     auto CopyEntireCells(wxString &s) {
         sys->clipboardcopy = s;
         wxString html = selected.g->ConvertToText(selected, 0, A_EXPHTMLT, this, false);
-        auto htmlobj = 
-        #ifdef __WXGTK__
-            new wxCustomDataObject(wxDF_HTML);
-            htmlobj->SetData(html.Len(), html);
-        #else
-            new wxHTMLDataObject(html);
-        #endif
+        auto htmlobj = new wxHTMLDataObject(html);
         return htmlobj;
     }
 
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1,6 +1,8 @@
 
 #include "stdafx.h"
 
+static_assert(wxCHECK_VERSION(3, 3, 0), "wxWidgets < 3.3.0 is not supported.");
+
 #ifndef __WXMSW__
 #define SIMPLERENDER
 #endif


aardappel:
Ah.. we'll also need a check that gives a compile error if someone tries to build with lower than 3.3.0 in that case (a static_assert)

src/document.h
```diff
@@ -413,12 +413,10 @@ struct Document {
...
         wxString html = selected.g->ConvertToText(selected, 0, A_EXPHTMLT, this, false);
-        auto htmlobj = 
-        #ifdef __WXGTK__
-            new wxCustomDataObject(wxDF_HTML);
-            htmlobj->SetData(html.Len(), html);
+        #if wxCHECK_VERSION(3, 3, 0)
+            auto htmlobj = new wxHTMLDataObject(html);
         #else
-            new wxHTMLDataObject(html);
+            static_assert(false, "wxWidgets < 3.3.0 is not supported.");
```
aardappel, line 419: why is this in this function? this can be at a global level. With the check inside the assert.

t2b3: Ok, I changed that. Is it ok now?

aardappel: Yup, perfect :)

aardappel:
Oh whoops, I merged, because CI succeeded, which means the 3.3.0 version is in their git repo?

t2b3:
> Oh whoops, I merged, because CI succeeded, which means the 3.3.0 version is in their git repo?

Well, in your CI workflow you instructed to use the git source of wxWidgets, so it will not hurt the compilation of the Linux binaries.

jn64:
3.3.0 is not released. AFAIK, right after the previous minor release (3.2.0), they change all the version strings in the master branch to 3.3.0, so it's not an indicator of whether 3.3.0 will be released soon (or at all).

[3.2.x is the stable branch of wxWidgets, while 3.3.x is the development branch.](https://www.wxwidgets.org/develop/roadmap/) Can I suggest, if it's possible, that TS use the stable branch? Some Linux distros will only package the stable branch, so if TS starts requiring wxWidgets >=3.3.x, then I won't be able to build my TS package for Fedora using Fedora's wxGTK package (I'll have to build wxWidgets from source).

Also, since wxWidgets 3.3.x doesn't actually exist yet, even bleeding edge distros like Arch can't package it. The [treesheets-git package in the Arch User Repository](https://aur.archlinux.org/packages/treesheets-git) would need to maintain a patch reverting this PR.

Of course, if you really require wxWidgets 3.3.x (or rather wxWidgets master branch) then I have no objection, do what's best for TS.

aardappel:
It's difficult to use the stable branch because wxWidgets even at the how old the project is remains remarkably buggy, they keep breaking, fixing and unfixing things which we respond to. We almost need to just pin commits to reliably work with them.

This distance between major releases is.. years? Will be hard for us to work with.

jn64:
Can we avoid a hard requirement on wxWidgets >=3.3.0, until 3.3.0 is actually released (or at least tagged)?

In other words, revert this PR as it was not intended to be merged until wxWidgets 3.3.0 released.

Besides Fedora/Arch packages mentioned, there are also Gentoo, NixOS, FreeBSD and other users who currently build TS with their distro wxWidgets 3.2.x that will be affected by this.

aardappel:
`3.2` was released 2023-02-08, `3.1` in February 29, 2016, about 7 years before.

`3.3` "There are no concrete plans for the next release in this branch yet."

I am not sure if we want to hang on to this patch for 7 years, waiting to apply it. If we decide we must be 3.2 compatible, then lets just nuke this commit, there is no point.

Do we honestly want to go that long without fixes?

This is a ridiculous development model.

jn64:
> Do we honestly want to go that long without fixes? This is a ridiculous
development model.

I'm not asking to use 3.2.0 for 7 years without fixes. The 3.2 branch is
actively developed, latest release is 3.2.2.1 (which is the version in
practically all distros).

Anyway do what's best for you; downstream users will adapt. Just noting that
this PR is a breaking change for some users.

t2b3:
> `3.2` was released 2023-02-08, `3.1` in February 29, 2016, about 7 years before.
> 
> `3.3` "There are no concrete plans for the next release in this branch yet."
> 
> I am not sure if we want to hang on to this patch for 7 years, waiting to apply it. If we decide we must be 3.2 compatible, then lets just nuke this commit, there is no point.
> 
> Do we honestly want to go that long without fixes?
> 
> This is a ridiculous development model.

Even when we stick to `3.2`, the function (adding a "custom" HTML data type into the composite data type to provide to clipboard) does not work properly because it truncates the string insite the custom data type. So this does not work properly on `3.2` and I agree with Wouter that the wxWidgets development circles are far too long.

aardappel:
@t2b3 do we have workaround for everything that is broken in 3.2? What currently happens if someone builds with 3.2?

t2b3:
> @t2b3 do we have workaround for everything that is broken in 3.2? What currently happens if someone builds with 3.2?

It will not compile due to the `static_assert`. If this is not active, it will build.

t2b3:
image
But given the hugh amount of commits in the meantime since the last release, I would like to stay closer to the git HEAD in order to not be surprised at the next release.

jn64:
All currently maintained packages I know of build TS with their respective distro wxWidgets 3.2.x:

- [Arch (AUR): treesheets-git](https://aur.archlinux.org/packages/treesheets-git)
- [Fedora (Copr): treesheets](https://copr.fedorainfracloud.org/coprs/jn64/treesheets/)
- [FreeBSD ports: treesheets](https://ports.freebsd.org/cgi/ports.cgi?query=treesheets&stype=all&sektion=all)
- [Nix: treesheets](https://search.nixos.org/packages?channel=unstable&type=packages&query=treesheets)

None build wxWidgets from master branch. I know this contradicts the build instructions so technically all the packages have been wrong for >1 year, but it's a matter of practicality. ~All distros package wxWidgets 3.2.x. Some distros will package 3.3.x when it's released, but it isn't yet; some won't.

Incidentally, some users might switch to the new Flatpak package and avoid this issue. But some users prefer distro packaging (why we aren't using upstream releases in the first place) so we'll just have to deal with it.

t2b3:
You can now use the Flatpak from Flathub
`flatpak install flathub com.strlen.TreeSheets`

aardappel:
@t2b3 I guess we maybe should undo this commit, and instead pin to the last 3.2.x release.
It sucks, but asking all these people to use a pinned commit is just not going to work I guess.

t2b3:
> @t2b3 I guess we maybe should undo this commit, and instead pin to the last 3.2.x release.
> 
> It sucks, but asking all these people to use a pinned commit is just not going to work I guess.

Ok I created a PR that does a switch between the clean instruction and the workaround depending on the version of wxWidgets on Linux. @aardappel Could you please take a look at it? Thanks.

PR #523, opened 2023-09-27
Add Flatpak manifest

t2b3:
This Pull Request adds the option to build a FlatPak and adds some conventions to fulfill the requirements of the Flatpak specification.

In order to successfully build and install it, wxWidgets has to be built and linked statically into TreeSheets but we can benefit from the newly added option to do exactly that.

The stuff related only to Linux is moved into `linux`, i.e. FreeDesktop related stuff.

This can be a step forward to move away from own pre-built binaries that create a lot of issues.


TS/org.aardappel.TreeSheets.metainfo.xml
```diff
@@ -0,0 +1,35 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
aardappel, line 1: Do all these metadata files have to sit in `TS` ? `TS` is also the distribution folder for other platforms, that clearly don't need this file.

aardappel:
@t2b3 and yes, if you want to maintain it, feel free to make a flatpak available

t2b3:
Hello Wouter
Thanks for your feedback. I removed the submodule and added instructions in the Flatpak manifest to extract the wxWidgets source into the folder `lib/wxWidgets`. In addition to that I moved all the Linux-only stuff out of `TS` and to the folder `linux` and changed the CMakefile to take the changes into account.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "lib/wxWidgets"]
+	path = lib/wxWidgets
```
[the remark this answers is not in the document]
t2b3: Reverted that changed.

TS/org.aardappel.TreeSheets.metainfo.xml
t2b3: I put the Linux-only stuff into a separate folder.

t2b3:
Hello Wouter
All checked passed. This PR would be good to go from my side.
I would like to ask you to review it and if it is ok, to merge it. Many thanks in advance!
Best regards
Tobias

t2b3:
Thanks for the merge :-)

PR #525, opened 2023-09-27
Conform to FlatHub requirements (2nd iteration)

t2b3:
See https://github.com/flathub/flathub/pull/4557 for more details.

This was needed to split into two rounds to get a proper release tag.


aardappel:
Seems like the CI error is unrelated to this change, will check

PR #534, opened 2023-10-05
Enable compile for released wxWidgets versions on Linux

t2b3:


src/main.cpp
```diff
@@ -1,8 +1,6 @@
 
 #include "stdafx.h"
 
-static_assert(wxCHECK_VERSION(3, 3, 0), "wxWidgets < 3.3.0 is not supported.");
```
aardappel, line 4: could make this 3.2?

t2b3: I pinned to the latest released version. Is this ok for you?

PR #537, opened 2023-10-08
Add option to Copy to clipboard with inline HTML images

t2b3:


aardappel:
Nice!

t2b3:
> Nice!

Thanks for your feedback.

PR #539, opened 2023-10-10
Add option to merge cell text

t2b3:
This adds a option to merge cells.

### DIFF
--- a/src/document.h
+++ b/src/document.h
@@ -914,11 +914,8 @@ struct Document {
         if (uk == WXK_NONE || (k < ' ' && k) || k == WXK_DELETE) {
             switch (k) {
                 case WXK_BACK:  // no menu shortcut available in wxwidgets
-                    if (!ctrl) { 
-                        return Action(dc, A_BACKSPACE);
-                    } else {  // prevent Ctrl+H from being treated as Backspace
-                        break;
-                    }
+                    if (!ctrl) return Action(dc, A_BACKSPACE);
+                    break; // Prevent Ctrl+H from being treated as Backspace
                 case WXK_RETURN: return Action(dc, shift ? A_ENTERGRID : A_ENTERCELL);
                 case WXK_ESCAPE:  // docs say it can be used as a menu accelerator, but it does not
                                   // trigger from there?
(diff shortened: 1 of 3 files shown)


aardappel:
Like I said in the other PR, this is a bit or a weird operation: "That is simpler, but also not so useful.. there is already a function to copy merged text, so the user can simply just use that to do the operation manually, copy the text, delete, and insert."

There's also the issue of UI clutter.. the Edit menu keeps growing, and now we have 2 operations that do almost the same thing, and are both quite specialized/niche.

t2b3:
> Like I said in the other PR, this is a bit or a weird operation: "That is simpler, but also not so useful.. there is already a function to copy merged text, so the user can simply just use that to do the operation manually, copy the text, delete, and insert."
> 
> 
> 
> There's also the issue of UI clutter.. the Edit menu keeps growing, and now we have 2 operations that do almost the same thing, and are both quite specialized/niche.

Well I do slightly disagree:
As you wrote, it would require three manual steps instead of just one as this proposal would also delete cells. I don't think it is not so niche as you might think. Just think of someone who pastes a multiline text that gets split up in TreeSheets but the user wants to keep it in one cell. I think this use case is rather common. Can you see my point?

aardappel:
Hmm, maybe we can try it, but lets use a clearer name for it, maybe "Collapse Cells", to make it clearer cells will be modified.
Also I don't see how your call to delete cells guarantees that the first cells remains? I guess because it contains text?

t2b3:
> Hmm, maybe we can try it, but lets use a clearer name for it, maybe "Collapse Cells", to make it clearer cells will be modified. Also I don't see how your call to delete cells guarantees that the first cells remains? I guess because it contains text?

Hello Wouter
Oh I missed that spot! I updated the PR to make sure that the first cell has always some content.
Thanks for your review :-) I also changed the name for the function as you proposed.
Kind regards
Tobias

src/document.h
```diff
@@ -1435,6 +1432,28 @@ struct Document {
                 }
                 return nullptr;
 
+            case A_COLLAPSE: {
+                if (selected.xs * selected.ys == 1) return _(L"More than one cell must be selected.");
+                auto fc = selected.GetFirst();
+                if (!fc->HasContent()) return _(L"The first cell in selection must have content.");
```
aardappel, line 1438: Why is this? even if this cell has no content, aren't you adding the text of all other cells to it below?
What I meant was I wasn't sure if `MultiCellDeleteSub` keeps this cell, but I guess it does?

t2b3: The problem arises when you have cells in the selection that all have no content and the selection is surrounded by cells without content. Then MultiCellDeleteSub will also delete the row/column of the first cell, leading to a crash...

t2b3: You are right we can check after collecting the text. If there is still no content, then leave the routine...

aardappel:
LGTM!

t2b3:
> LGTM!

Merci :-)

PR #540, opened 2023-10-12
Workaround is no longer needed

t2b3:


aardappel:
@jn64 since this the same change as before that prompted the change to 3.3, do you have a sense how much trouble we're causing for these distros by pushing forward the 3.2.x requirement instead that this PR does?

jn64:
No trouble whatsoever, because 3.2.3 is an actual release (unlike 3.3), and it's a minor release in the current stable branch, so we can reasonably expect distros to update in a timely manner. Arch already has it, Fedora might take a few days to a week for testing, not sure about the others.

If there are any problems or delays on the distro side, e.g. for whatever reason Fedora doesn't update to 3.2.3, that's a problem for Fedora (and me), not you. Even if the outcome is the same as what I mentioned regarding 3.3 (say I end up patching my TS package to revert this change), my complaint would only be with Fedora. I would not ask you to hold back this change.

---

Though if I may request one thing, please mention such changes in the merge/commit message (e.g. "Bump wxWidgets to 3.2.3") to make it easier to spot, not just for packagers but anyone building TS.

aardappel:
Ok, makes sense, thanks!

PR #542, opened 2023-10-14
Tune the input for the hash

t2b3:
Profiling TreeSheets shows that hashing the entire image data takes up a lot of time.

This commit tweaks this by just hashing a specific chunk of the image data at its beginning.

### DIFF
--- a/src/system.h
+++ b/src/system.h
@@ -577,7 +577,9 @@ struct System {
     }
 
     int AddImageToList(double sc, vector<uint8_t> &&idv, char iti) {
-        auto hash = FNV1A64(idv);
+        int max = 4096;
+        auto subvector = vector<uint8_t>(idv.begin(), idv.begin() + min(idv.size(), max));
+        auto hash = FNV1A64(subvector);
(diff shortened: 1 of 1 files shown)


AntonBogun:
This may be late (I am on my own separate fork), but I looked at it and wouldn't this mean that if you have an image that has a portion of it empty in the beginning (i.e. transparency in icons), it would take up the hash slot for all other images that have that same initial portion?
For a 512x512 image, if the first 8 rows of two images are the same this would treat them as the same, which is probably not intended.

aardappel:
Agree this is a potential problem, but having many images of the sizes as large as 512x512 start the same is not super common? If you have many similar images, typically they're smaller.

And what happens when you have a hash collision? Do we only do full comparisons when you load a new image, or what?

This code could be improved by hashing bytes from multiple locations in the image. But now this may be cache-inefficient so you'd have to make sure to has 64 aligned bytes at a time, etc :)

Or for simplicity, maybe just read the middle of the image, as being the least likely place to be similar.

PR #543, opened 2023-10-15
Iterate only through the cells array in grid to set selection

t2b3:


src/document.h
```diff
@@ -139,21 +139,15 @@ struct Document {
...
-                foreachcellingrid(c, ipg) {
-                    if (c == ics) {
-                        SetSelect(Selection(ipg, x, y, xs, ys));
-                        goto cellisselected;
-                    }
-                }
+        if (Grid *ipg; ics && (ipg = ics->parent->grid)) {
+            loop(i, ipg->xs * ipg->ys) if (ipg->cells[i] == ics) {
+                SetSelect(Selection(ipg, i % ipg->xs, i / ipg->xs, xs, ys));
+                break;
```
aardappel, line 145: Ok, took me a while to figure out what you have done here. Essentially, you are able to replace a goto by a break, but the cost of that is a hand-coded loop that is hard to read. You have to understand what this loop does and how it is different from all uses of `foreachcellingrid` in the code. The answer is that it is not different, just less loops. I think this makes the code worse.

If you really wanted this, I'd put it in a macro `foreachcellingrid_singleloop` or whatever, to at least centralize that it does the same thing as before. But I would not do that either.

Instead, if you dislike the goto, one small improvement is to move the goto label to directly after the `foreachcellingrid`, which is much more readable.

aardappel: Or, if you wanted to get really fancy in avoiding goto, you could make all `foreachcellingrid` automatically emit a `foreachcellingrid_end:` label, then define a `break_foreachcellingrid` macro that does the goto.. that way you've standardized how to break out of these double loops.

t2b3: Hello Wouter
Taking your comments into consideration it might be better to leave it as is. Putting the goto label after the loop gives a compiler warning because the compound statements ends there. Modifying the macro to emit that label at the end does not seem to be fit for macros.

t2b3: On the other side, the loop macro emits also x and ys but looking through the code, this is the only occurrence where x and y are directly used?

aardappel: You can get rid of that warning simply by adding `;` right after the label.

I guess? A macro can do anything :)

t2b3: Ok, I rearranged that code for better readability. But I refrain from adding something to the macro because this seems to be too complicated for me.

t2b3: See the other PR then.

PR #546, opened 2023-10-16
Get rid of Focus check entirely

t2b3:
Only activation events for the application are filtered with this commit.


t2b3:
> Ok, I don't know what these changes do, but I presume you have tested they work on win/lin at least :)

I did. Thanks for the merge!

Issue #562, opened 2023-12-06
Crashes in `FindNextSearchMatch` when called from `OnTabChange`


aardappel:
I've seen this crash log a few times now:

```
C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.37.32822\include\xstring (2492): std::basic_string<wchar_t,std::char_traits<wchar_t>,std::allocator<wchar_t> >::basic_string<wchar_t,std::char_traits<wchar_t>,std::allocator<wchar_t> >
C:\w\treesheets\src\cell.h (416): treesheets::Cell::FindNextSearchMatch
C:\w\treesheets\src\document.h (2020): treesheets::Document::SearchNext
C:\w\treesheets\src\myframe.h (826): treesheets::MyFrame::OnTabChange
```

It's an illegal access: `ExpCode: 0xC0000005 ExpFlags: 0 ExpAddress: 0x69D4405B`

I can't see from the code why it would do this, but it looks like some part of the state is not set up correctly when switching tabs.

t2b3:
I tried to reproduce the issue on Linux (wxGTK) but it does not seem to be a problem there. I will try it out the next time I cross a Windows installation.

PR #563, opened 2023-12-06
Constify search string

t2b3:
Maybe not directly related to https://github.com/aardappel/treesheets/issues/562, but it is maybe still an improvement?


aardappel:
Can't hurt :)

PR #573, opened 2023-12-21
Constify read-only Selection argument for FindLink and FindExact

t2b3:
This commit constifies the Selection argument passed by reference for
`{Cell,Grid}::{FindLink,FindExact}(...)` as they are only read.

### DIFF
--- a/src/cell.h
+++ b/src/cell.h
@@ -438,7 +438,7 @@ struct Cell {
         return best;
     }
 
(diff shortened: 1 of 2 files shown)


aardappel:
const changes are so simple and harmless, you can probably combine some in one pr. Or rather, just push it yourself, I can't imagine `&` to `const &` ever being a problematic change if it all still compiles :)

PR #574, opened 2023-12-23
Simplify, improve and enhance replace logic

t2b3:
This commit allows to replace a search text with an empty replace string, so to delete the match text.

It also removes a bug where the matched string was replaced twice in the case-insensitive mode.

### DIFF
--- a/src/text.h
+++ b/src/text.h
@@ -423,32 +423,25 @@ struct Text {
     }
 
     void ReplaceStr(const wxString &str, const wxString &lstr) {
-        wxString lowert;
-        wxString *fort;
-        const wxString *istr;
-        
-        if (lstr.IsEmpty()) {
-            fort = &t;
-            istr = &str;
-        } else {
-            lowert = t.Lower();
-            fort = &lowert;
-            istr = &lstr;
-        }
-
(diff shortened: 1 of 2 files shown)


src/text.h
```diff
@@ -422,33 +422,27 @@ struct Text {
...
-
-        for (int i = 0, j; (j = fort->Mid(i).Find(sys->searchstring)) >= 0;) 
-        {
-            // does this need WasEdited()?
-            i += j;
-            fort->Remove(i, sys->searchstring.Len());
-            fort->insert(i, *istr);
-
-            if (!sys->casesensitivesearch) {
+            wxString lowert = t.Lower();
```
aardappel, line 435: `Lower()` is an expensive operation, since it allocates a new string and does a unicode sensitive lowercase operation on the whole string.. that should ideally not be done repeatedly.

t2b3: Hello Wouter,
Thanks for your review. This is unavoidable according to my opinion for `lowert` but for `lstr` it is not, so I restored the currently used logic. Is this ok now?
Kind regards
Tobias

aardappel: yeah it is unfortunate wxWidgets doesn't have a case insensitive find, that would at least avoid the allocation.

t2b3: Thanks for your merge!

PR #581, opened 2024-01-06
Implement lazy search

t2b3:
If no search term is given and one cell is selected and the user triggers the search, use the cell text string as search query.

This is similiar to the F6 and F7 (jump to text/image) experience.

Also make the rootgrid check the first check for futureproofness.

### DIFF
--- a/src/document.h
+++ b/src/document.h
@@ -1175,12 +1175,17 @@ struct Document {
                 return nullptr;
             }
 
-            case A_SEARCHNEXT: {
-                return SearchNext(dc, false, true, false);
-            }
-
+            case A_SEARCHNEXT: 
             case A_SEARCHPREV: {
-                return SearchNext(dc, false, true, true);
+                if (sys->searchstring.Len()) return SearchNext(dc, false, true, k == A_SEARCHPREV);
+                if (Cell *c = selected.GetCell()) {
+                    if (!c->text.t.Len()) return _(L"No text in this cell.");
+                    sys->frame->filter->SetFocus();
+                    sys->frame->filter->SetValue(c->text.t);
+                    return nullptr;
+                } else {
+                    return _(L"You need to select one cell if you want to search for its text.");
+                }
             }
 
             case A_CASESENSITIVESEARCH: {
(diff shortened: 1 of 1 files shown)


src/document.h
```diff
@@ -2024,10 +2023,21 @@ struct Document {
         }
     }
 
+    const wxChar *LoadSearch() {
```
aardappel, line 2026: Unless they're gigantic, I'd generally say there is no good reason introducing functions that only get called once, makes the code harder to follow

src/document.h
```diff
@@ -2024,10 +2023,21 @@ struct Document {
         }
     }
 
+    const wxChar *LoadSearch() {
+        if (Cell *c = selected.GetCell()) {
+            if (!c->text.t.Len()) return _(L"No text in this cell.");
+            sys->frame->filter->SetFocus();
+            sys->frame->filter->SetValue(c->text.t);
+            return _(L"Search string was loaded from selection.");
```
aardappel, line 2031: This status is maybe superfluous, status should only be set on special events

src/document.h
```diff
@@ -1175,12 +1175,11 @@ struct Document {
...
-            case A_SEARCHNEXT: {
-                return SearchNext(dc, false, true, false);
-            }
-
+            case A_SEARCHNEXT: 
             case A_SEARCHPREV: {
-                return SearchNext(dc, false, true, true);
+                return sys->searchstring.Len()
+                    ? SearchNext(dc, false, true, k == A_SEARCHPREV)
+                    : LoadSearch();
```
aardappel, line 1182: this function only sets the search, so does that mean I have to press search twice to do the actual search? what is the idea behind that UI (as opposed to searching directly?)

src/document.h
t2b3: > This status is maybe superfluous, status should only be set on special events

Ok, I was unsure myself. I replaced it with `nullptr`.

src/document.h
t2b3: Ok, I changed that.

src/document.h
t2b3: I think of two reasons:
1. First provide some kind of visual feedback that the search is now active (the cells are darkened) and ready to proceed.
2. The search parameters are set by `MyFrame::OnSearch` that is run when the value of the `wxTextCtrl` `filter` is changed. So this runs asynchronously and we cannot just run `SearchNext` without knowing for sure that the search parameters are all set.

aardappel: Ok, I suppose that's fine for the moment.. much better than not working at all :)

t2b3: Thanks for your review!

PR #583, opened 2024-01-06
Allow also to search for selected text

t2b3:
Instead of using only the text of the whole cell, allow also to use only the part of cell text that is selected.


t2b3:
Thanks for the merge!

PR #585, opened 2024-01-07
Give user more control over time

t2b3:
1. Allow user to filter for exact datetime ranges.
2. Show user in status bar also the exact time of the last modification.


t2b3:
Thanks for your review!

PR #589, opened 2024-01-09
Place datetime filter dialog on stack

t2b3:
This code is cleaner.

In addition add also some border around the elements. This looks nicer especially on the border of the dialog.

### DIFF
--- a/src/document.h
+++ b/src/document.h
@@ -1262,56 +1262,56 @@ struct Document {
                                          : _(L"1:1 scale restored.");
 
             case A_FILTERRANGE: {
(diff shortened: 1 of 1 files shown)


aardappel:
very nice!
was this a memory leak before? I don't see them being deleted in the old code..

t2b3:
> very nice! was this a memory leak before? I don't see them being deleted in the old code..
I am testing it out now. I mean that I did not see a memory leak before. But as it is possible for modal dialogs to create it on the stack, I would rather prefer it anyway. :-)

PR #591, opened 2024-01-10
Do not copy folded grids to clipboard

t2b3:


aardappel:
So now "copy" does not copy folded cells but "export" does?
That might be confusing to the user.
What is the reason behind not wanting to copy folded cells?

t2b3:
The main reason for me is the expectation of the user: If the user has a folded grid, he does not see the folded grid in the UI. So when he copies to the clipboard, it should be consistent with what is displayed. If the user also wants to copy the folded grids, the user may unfold it. Do you agree? With export it is maybe more the expectation that it should include everything.

aardappel:
I'm not sure if I agree that is what the user would expect. A user could expect either one, and not getting the folded cells when you expect them seems more problematic than getting the folded cells when you don't expect them. And then having no way to override that behavior.

If you get folded cells you didn't want you can delete them, if you are missing your folded cells you wanted you have to do the action again. Or if you don't notice, you now have a bad copy.

Did anyone complain about this behavior?

We could have a way to choose this behavior, but that is unnecessary complex UI.

Seems what we had before was on average better.

t2b3:
> I'm not sure if I agree that is what the user would expect. A user could expect either one, and not getting the folded cells when you expect them seems more problematic than getting the folded cells when you don't expect them. And then having no way to override that behavior.
> 
> If you get folded cells you didn't want you can delete them, if you are missing your folded cells you wanted you have to do the action again. Or if you don't notice, you now have a bad copy.
> 
> Did anyone complain about this behavior?
> 
> We could have a way to choose this behavior, but that is unnecessary complex UI.
> 
> Seems what we had before was on average better.

Ok, thanks for your reply.
I can understand the reasoning and I will close this PR then.

PR #593, opened 2024-01-14
`wxSizeEvent` should not trigger a repaint of the document

t2b3:
The document is rendered into a scrolled window. If the size of the scrolled window for the canvas is changed, the rendered document can stay the same as it can be navigated by the scrollbars.

This should avoid the expensive `Document::Refresh()` call.


aardappel:
But if it gets bigger, what causes the new area to be rendered? I presume that still happens?

PR #594, opened 2024-01-14
Avoid extraneous call for `DrawSelect` in case of a zoom

t2b3:
`Document::Zoom(int, wxDC&, bool, bool)` calls
`Document::DrawSelectMove(wxDC&, Selection&, bool, bool)` which calls `Document::DrawSelect(wxDC&, Selection&, bool, bool)`. So in this case, we do not need to call
`Document::DrawSelect(wxDC&, Selection&, bool, bool)` again.

### DIFF
--- a/src/document.h
+++ b/src/document.h
@@ -667,17 +667,18 @@ struct Document {
         sw->DoPrepareDC(dc);
         ShiftToCenter(dc);
         Render(dc);
-        DrawSelect(dc, selected);
-        if (hover.g) hover.g->DrawHover(this, dc, hover);
-        if (scaledviewingmode) { dc.SetUserScale(1, 1); }
         if (initialzoomlevel) {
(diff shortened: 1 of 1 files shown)


aardappel:
Are you aware that at least on Windows the selection is drawn as a XOR operation (i.e. invert pixels), meaning that drawing it a second time undoes the drawing? Which is faster than redrawing the entire screen.

t2b3:
> Are you aware that at least on Windows the selection is drawn as a XOR operation (i.e. invert pixels), meaning that drawing it a second time undoes the drawing? Which is faster than redrawing the entire screen.

I am aware of this, yes. Unfortunately we have to stick with `SIMPLERENDER` on Unix platforms.

t2b3:
> So does this change work well on Windows?

I tested it on Windows and it works well.

PR #598, opened 2024-01-14
For the initial zoom do not call Refresh again

t2b3:


t2b3:
https://github.com/aardappel/treesheets/pull/598
https://github.com/aardappel/treesheets/pull/595
https://github.com/aardappel/treesheets/pull/594
should make `Refresh()` only called once when document is loaded.

PR #599, opened 2024-01-17
In case of `initialzoom` set, set up `drawpath` first and then do the…

t2b3:
… layout and render

This commit saves the extra layouting and rendering before the zoom.

### DIFF
--- a/src/document.h
+++ b/src/document.h
@@ -512,7 +512,7 @@ struct Document {
         return;
     }
 
-    void Zoom(int dir, wxDC &dc, bool fromroot = false, bool selectionmaybedrawroot = true, bool needsrefresh = true) {
+    void ZoomSetDrawPath(int dir, wxDC &dc, bool fromroot = true, bool selectionmaybedrawroot = true) {
         int len = max(0, (fromroot ? 0 : drawpath.size()) + dir);
         if (!len && !drawpath.size()) return;
         if (dir > 0) {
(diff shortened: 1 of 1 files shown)


src/document.h
```diff
@@ -525,6 +525,7 @@ struct Document {
                 SetSelect(drawroot->parent->grid->FindCell(drawroot));
         }
         while (len < drawpath.size()) drawpath.remove(0);
+        if (onlysetupdrawpath) return;
```
aardappel, line 528: In how many places is `Zoom` called? if only few, it may be cleaner than rather using a boolean argument, to just split this function in 2 (e.g. `ZoomSetDrawPath` and `ZoomDraw`) and call those as needed.

Just an idea.. boolean args to a function that make it take different paths thru the function is usually not so great, often better to split or specialize the function for the different callers instead

aardappel (commented):
Very cool all these performance optimizations.. looks like you're starting to understand all the code pretty deeply now :)

src/document.h
t2b3: `Zoom` is called from five places in the code. In every instance,  we have a complete layout calculation and the draw of the selection. So currently it makes sense to split out the setting of the `drawpath` for the initial zooming but not to move the other part of the `Zoom` function into a second separate split function. Do you agree?
Thanks for your feedback in advance!

t2b3:
> Very cool all these performance optimizations.. looks like you're starting to understand all the code pretty deeply now :)

It takes some time, step after step I learn a little bit more. And I try to learn from my mistakes. ;-)

src/document.h
t2b3: See the updated commit in the PR :-)

aardappel: sure.

src/document.h
```diff
@@ -525,6 +525,10 @@ struct Document {
                 SetSelect(drawroot->parent->grid->FindCell(drawroot));
         }
         while (len < drawpath.size()) drawpath.remove(0);
+    }
+
+    void Zoom(int dir, wxDC &dc, bool fromroot = false, bool selectionmaybedrawroot = true) {
```
aardappel, line 530: `selectionmaybedrawroot` not needed anymore?

PR #600, opened 2024-01-18
Do not set the status text with search result on tab change

t2b3:


src/system.h
```diff
@@ -158,7 +158,7 @@ struct System {
         newdoc->sw->SetFocus();
         newdoc->UpdateFileName();
         wxClientDC dc(newdoc->sw);
-        newdoc->sw->Status(newdoc->SearchNext(dc, false, false, false));
+        (void) newdoc->SearchNext(dc, false, false, false);
```
aardappel, line 161: you don't need a `(void)` there, as we don't do that anywhere else where we ignore return values. And if you did, it would be without a space.

PR #601, opened 2024-01-19
Wrap as you type on multiple cell selection

t2b3:


src/document.h
```diff
@@ -972,10 +972,13 @@ struct Document {
         } else if (uk >= ' ') {
             if (!selected.g) return NoSel();
             Cell *c = selected.ThinExpand(this);
-            if (!c) return OneCell();
+            if (!c) {
+                selected.Wrap(this);
```
aardappel, line 976: Does this always result in a single cell?

t2b3: In `Selection::Wrap(Document*)` there is `xs = ys = 1` so I assume yes.

t2b3:
Thanks for your review!

PR #603, opened 2024-01-25
Fix crash when moving cells after copy paste while in textedit

AntonBogun:


t2b3:
Hi @AntonBogun 
Could you describe how you produce the crash? Thanks in advance!
Best regards,
Tobias

AntonBogun:
Produce the crash like this:
Write some text in two cells (unsure if necessary but regardless),
Select them
Move over to a different cell and enter text edit
Paste (you should be selecting cells at this stage except treesheets didn't realize you're not in text edit anymore)
Finally, press ctrl+right (or any arrow I think)

t2b3:
Hi @AntonBogun

Thanks for your reply. I could reproduce the issue. Thanks for pointing out to this issue.

While I definitively see your valid idea in your PR, I have two major concerns with this PR: 

1. While its intention seems right to me, I think that it is "too late" to switch the textedit boolean at the proposed place. Rather than letting the wrong state of `Selection::textedit` get into the function `void Selection::Dir(Document*, bool, bool, wxDC&, int, int, int, int, int, bool, bool, bool)` and correct by an extra condition check (`GetCell()`), it should be fixed at the origin of the bug. According to my investigation the bug roots in the miss of the `textedit` not being switched off when a grid gets merged into its parent grid (`void Grid::MergeWithParent(Grid*, Selection&)`), so I created another proposal in a PR.
2. It is quiet expensive to check the condition `GetCell()` at every call of `void Selection::Dir(Document*, bool, bool, wxDC&, int, int, int, int, int, bool, bool, bool)`. As said in 1, we should come "clean" to the function.

I hope that you can agree with my points here.
@aardappel 
Which way do you prefer? My alternative approach lies in PR https://github.com/aardappel/treesheets/pull/604

Best regards,
Tobias

PR #605, opened 2024-01-27
Replace toolbar icons with scalable symbols from Material Design

t2b3:
The SVG files are rendered into bitmap bundles that can be used for HiDPI display.

This commit also removes the extra icon set for black and white and its configuration. The Material Design symbols are already monochrome.


aardappel:
Nice! With visual changes like this though, can you post a preview screenshot?
Does this rendering work on all 3 platforms?

t2b3:
On Linux for example:

image

aardappel:
Ok cool, let's try it out :)

PR #612, opened 2024-03-13
Do not install wxWidgets along with TreeSheets static build

t2b3:
This avoids that a install and uninstall target for wxWidgets itself is created. With this considered, only the TreeSheets files are installed, but none from wxWidgets.


t2b3:
Grazie!🙏

PR #613, opened 2024-03-23
Add selection cycle

t2b3:
Jump to the opposite side of the grid at the grid border

### DIFF
--- a/src/selection.h
+++ b/src/selection.h
@@ -286,6 +286,20 @@ class Selection {
                 {
                     if (dx + dy < 0) v--;
                     vs = 1;  // make it a cell selection
+                } else { // selection cycle, jump to the opposite side of the grid
+                    if (y + dy > g->ys) {
+                        y = 0;
+                        vs = 1;
+                    } else if (y + dy < 0) {
+                        y = g->ys - 1;
+                        vs = 1;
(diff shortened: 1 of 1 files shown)


t2b3:
https://github.com/aardappel/treesheets/assets/3286756/9824fb20-ec4d-44ba-9c43-18f09762b09c

aardappel:
I am not sure if I'd agree this is a good feature. I very often hold down a direction key to get to the edges of a grid, so I can insert new cells there. I am not sure that being able to wrap around is more useful?

t2b3:
> I am not sure if I'd agree this is a good feature. I very often hold down a direction key to get to the edges of a grid, so I can insert new cells there. I am not sure that being able to wrap around is more useful?

Well, you can still add the new cells because the grid edge needs to be first selected before a wrap can happen. Only when you are in thin selection of the edge and then press the key to leave the grid, then it is wrapped.

aardappel:
Ah I didn't understand that it worked that way from the code ;)
Ok, that could be an improvement then, lets try it out!

t2b3:
> Ah I didn't understand that it worked that way from the code ;)
> 
> Ok, that could be an improvement then, lets try it out!

Thanks for your reply and the merge!

PR #616, opened 2024-03-28
Replace wxScrolledWindow with wxScrolledCanvas

t2b3:
According to <https://docs.wxwidgets.org/3.2.5/classwx_scrolled.html>, `wxScrolledCanvas` is more suited to the use case of TreeSheets as there are no child controls in `TSCanvas`.


aardappel:
Hmm, I wonder if there are any behavioral differences? I guess let's try it :)

t2b3:
Happy Easter!

PR #617, opened 2024-03-29
Put clipboard data on the system stack

t2b3:
The clipboard data structures are created by the first document on the heap and then are subsequently overridden by other documents.

It can be a better design to put the clipboard data into one central place on the stack of `System` as it is available during the lifetime of the program.


aardappel:
nice, much cleaner!

PR #618, opened 2024-03-29
Initialize clipboard data structures

t2b3:
Always a good idea


aardappel:
I believe this is redundant, it will always call the default constructor.

t2b3:
> I believe this is redundant, it will always call the default constructor.

Ah ok :-) Thanks for the clarification.

PR #619, opened 2024-03-30
Avoid wxDataObjectComposite for pasting

t2b3:
wxDataObjectComposite does not work properly on Wayland for text atoms, so avoid it.

In addition to that, the paste operation wxDataObjects for text, image and file name can be handled completely on the stack.

Only the drag and drop operation still requires an composite data object on the heap, so handle it separately.

### DIFF
--- a/src/document.h
+++ b/src/document.h
@@ -2103,12 +2103,13 @@ struct Document {
         Cell *c = selected.ThinExpand(this);
         if (!c) return;
         wxBusyCursor wait;
+        bool wantsrefresh = false;
 
         if (pdataobjt.GetText() != wxEmptyString) {
             wxString s = pdataobjt.GetText();
             if ((sys->clipboardcopy == s) && sys->cellclipboard) {
                 c->Paste(this, sys->cellclipboard.get(), selected);
-                Refresh();
+                wantsrefresh = true;
             } else {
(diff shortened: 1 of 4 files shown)


t2b3:
For details, please see
https://github.com/wxWidgets/wxWidgets/issues/24444

src/document.h
```diff
@@ -2083,75 +2088,55 @@ struct Document {
...
+                     const wxBitmapDataObject &pdataobji,
+                     const wxFileDataObject &pdataobjf) {
         Cell *c = selected.ThinExpand(this);
         if (!c) return;
         wxBusyCursor wait;
-        switch (sys->dataobjc->GetReceivedFormat().GetType()) {
-            case wxDF_FILENAME: {
-                const wxArrayString &as = sys->dataobjf->GetFilenames();
+
+        if (pdataobjt.GetText() != wxEmptyString) {
```
aardappel, line 2098: maybe add some comments to the code explaining why we do things the way we do? why certain things are necessary on what platform? So we don't accidentally change it back in the future?

aardappel: you add a detail to the commit and the wxWidgets bug, but for long term knowledge, its best to be in the code.

t2b3: Hello Wouter
Thanks for your feedback. Please see the commit added. Would this be ok for you?
Thanks for your reply in advance!

PR #620, opened 2024-04-04
Keep track of images undergoing manipulation

t2b3:
Prior to this commit, the image operation got executed on each image in the cells of the selection *per cell*.

The issue was that if the same image was referenced in multiple cells in the selection, the operation was executed per cell, thus on the same image multiple times.

In order to prevent this, keep track of all images and the cells referring to them in the selection, to execute the image operation only once on each image and to reset the layout of the referencing cells.

### DIFF
--- a/src/document.h
+++ b/src/document.h
@@ -1789,6 +1789,7 @@ struct Document {
                 }
 
                 for(auto c: cellstoresetlayout) {
+                    c->ResetChildren();
                     c->ResetLayout();
                 }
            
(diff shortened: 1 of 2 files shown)


aardappel:
Nice, good fix!

t2b3:
> Nice, good fix!

Thanks for the kind feedback!

t2b3:
I added some small second commit to reset also children because then it is ensured that the layout is updated to the image dimensions in every case. 

From my side, the PR would be ready to go. :-)

t2b3:
Merci and have a nice day!
Beste Grüsse aus Olten
Tobias

PR #622, opened 2024-04-06
Apply changes from clang-format

t2b3:


src/myapp.h
```diff
@@ -18,79 +18,77 @@ struct MyApp : wxApp {
...
-        #if wxUSE_UNICODE == 0
-        #error "must use unicode version of wx libs to ensure data integrity of .cts files"
-        #endif
+#if wxUSE_UNICODE == 0
+    #error "must use unicode version of wx libs to ensure data integrity of .cts files"
+#endif
         ASSERT(wxUSE_UNICODE);
 
-        #ifdef __WXMAC__
+#ifdef __WXMAC__
```
aardappel, line 40: We definitely want ifdefs indented.. if there are no settings for clang format that can fix this, we either have to fix all ifdefs manually afterwards (like I did last time I think), or not apply clang format.

t2b3: Hi Wouter
Thanks for your feedback. I also did first a automatic round and then picking the right ones with `git add --patch`. Please see the series of commits.
Best regards and a nice sunday
Tobias

aardappel:
Thanks! Not all of these are better than what they replace.. but they have the advantage that we don't need to reconsider them in future clang-formats, so it's ok to merge.

PR #626, opened 2024-04-10
wxAUI enhancements

t2b3:


src/myframe.h
```diff
@@ -3,14 +3,14 @@ struct MyFrame : wxFrame {
     wxString exepath_;
     wxFileHistory filehistory;
     wxTextCtrl *filter, *replaces;
-    wxToolBar *tb;
+    wxAuiToolBar *tb1, *tb2, *tb3, *tb4, *tb5, *tb6, *tb7, *tb8, *tb9;
```
aardappel, line 6: give these an actual name of whats in them?

t2b3: :-) Oh, you are completely right! I corrected this and pushed an additional commit for it in this PR.

aardappel:
Looks cool.. wonder what it looks like..

PR #627, opened 2024-04-12

georgeraraujo:
https://github.com/aardappel/treesheets/pull/622 reverted part of what https://github.com/aardappel/treesheets/pull/281 did, so do it again. The rationale is [here](https://github.com/aardappel/treesheets/pull/281#issuecomment-1308850604).

Fixed as described in the [wikidump of the Audacity Wiki as of March 31, 2023 that can be found in the Internet Archive](https://web.archive.org/web/20221112235157/https://wiki.audacityteam.org/wiki/Translating_Audacity/Writing_Translatable_Code#Dealing_with_Long_Strings).


aardappel:
Ahh.. yeah that is an unfortunate side effect of clang-format.. wish we could turn that off.
@t2b3

t2b3:
I currently see no convenient way:
https://clang.llvm.org/docs/ClangFormatStyleOptions.html

One option would be to rescind the column limit but maybe not worth because the side effects.

t2b3:
Thanks for the quick fix and sorry for having missed out that side effects with the translateable strings.

PR #629, opened 2024-04-26
Comment is no longer valid

tobiolo:
Issue solved with https://github.com/wxWidgets/wxWidgets/pull/24505

But it makes sense to keep the changes and not to pack the data objects into a composite object only to unpack it later.

### DIFF
--- a/src/document.h
+++ b/src/document.h
@@ -1566,17 +1566,13 @@ struct Document {
             case wxID_PASTE:
                 if (!(c = selected.ThinExpand(this))) return OneCell();
                 if (wxTheClipboard->Open()) {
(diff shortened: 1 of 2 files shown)


aardappel:
These kinds of fixes are not available in the minimum wxWidgets version we support yet, right? So some people building or even looking at the code may still have the problem. Thus, to be totally correct, it be better to say that this is a problem which was fixed in $date version of wxWidgets. Only when we require a minimum version of wxWidgets to build against that does not have the bug should the code pretend there is no problem.

tobiolo:
Thanks for your feedback. Then it is really better to wait for the next release version. I would like to keep this PR open as a reminder then.

PR #634, opened 2024-04-29
Member variables for selection position and length are never negative

tobiolo:
Use uint8_t. `wxDataOutputStream::Write8(wxUint8)` and `wxDataInputStream::Read8` also uses only unsigned integers.

### DIFF
--- a/src/document.h
+++ b/src/document.h
@@ -132,7 +132,7 @@ struct Document {
 
     uint Background() { return rootgrid ? rootgrid->cellcolor : 0xFFFFFF; }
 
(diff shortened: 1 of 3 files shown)


aardappel:
Something being never negative is a bad reason for using `unsigned` in C++. You make code more brittle since there are lots of subtle code patterns where substracting can now wrap (which usually has worse results than being negative). Also mixing signed and unsigned can produce subtle issues. Best is to simply use signed everywhere.

tobiolo:
Thanks for your explanation! I came to the same conclusion then, but not with the perspective of the wrap in subtraction cases. Thanks for having pointed out!

PR #635, opened 2024-04-29
Add feature: bookmark

tobiolo:
With this feature, users can bookmark cells, delete these bookmarks and go to bookmarks. Bookmarks are persistent.

### DIFF
--- a/src/cell.h
+++ b/src/cell.h
@@ -321,21 +321,22 @@ struct Cell {
         doc->AddUndo(this);
     }
 
-    void Save(wxDataOutputStream &dos, Cell *ocs) const {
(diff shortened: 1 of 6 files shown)


tobiolo:
Still wonder what would be better:
expand Cell with bookmarked member variable or
store Cells as set centrally?

tobiolo:
Hello Wouter,
Thanks for your reply. Thanks for outlining possible ways to implement it and especially for challenging the PR. I see your point and having a second thought about it, I see that it is sufficient to organise by the structure in itself and not having a parallel organization by means of bookmarks. There is also the link feature so this is also sufficient and an alternative way to have an equivalent way of maintaining a sort of bookmark. So I withdraw my PR.
Best regards,
Tobias

PR #636, opened 2024-04-29
Adjust to C++20

tobiolo:
This series of commits updates the toolchain and also the code so C++20 is used.


aardappel:
Awesome! Progress!

PR #637, opened 2024-04-30
Replace replace all icon

tobiolo:
Proposal for a replacement:
It is like the replace icon but within the document icon.


aardappel:
Neither is great, the old one looks like "search again", and the new one is hard to even see what it is. So I don't mind either one, your call.

Not sure what would be better.. I don't see any good icons online.

tobiolo:
> Neither is great, the old one looks like "search again", and the new one is hard to even see what it is. So I don't mind either one, your call.
> 
> 
> 
> Not sure what would be better.. I don't see any good icons online.

You are right. Better try something more readable!

georgeraraujo:
What do you think of https://github.com/microsoft/vscode-icons/blob/main/icons/light/replace-all.svg ?

As far as I can tell, the license permits use in other projects: https://github.com/microsoft/vscode-icons/blob/main/LICENSE

aardappel:
I guess that is pretty good.. looks a bit like the one that used to be in visual studio I think?

georgeraraujo:
> I guess that is pretty good.. looks a bit like the one that used to be in visual studio I think?

It is used in Visual Studio Code, so it could be.

georgeraraujo:
I replaced the current icons with the ones from https://github.com/microsoft/vscode-icons/ - I modified them to be all black for more contrast, and created edit/copy and edit/paste from existing icons:

image

What do you think?

aardappel:
Those icons are pretty good.. but maybe we can use a version with slighty thinner lines?

My VSCode looks like:

image

georgeraraujo:
> Those icons are pretty good.. but maybe we can use a version with slighty thinner lines?
> 
> My VSCode looks like:
> 
> image

I think the lines in my screenshot look thicker because I changed the grayish color in the stock VSCode icons to black. Here's how TreeSheets looks like with the stock VSCode icons (except for Paste, New Grid, and Image, for which VSCode had no corresponding icons):

image

aardappel:
By changing them to black you also changed the edge pixels which presumably had an alpha component, making it look thicker.

So yeah, thinner looks better, but you'd need icons for Paste etc that look the same.

georgeraraujo:
> By changing them to black you also changed the edge pixels which presumably had an alpha component, making it look thicker.
> 
> So yeah, thinner looks better, but you'd need icons for Paste etc that look the same.

Here's what I came up with:

image

aardappel:
looks nice to me, wdyt @tobiolo

tobiolo:
Hi @georgeraraujo, @aardappel 
I like that icon set very much. Thus I would love to see a PR landing. 😄 Thanks for your work @georgeraraujo

georgeraraujo:
> Hi @georgeraraujo, @aardappel I like that icon set very much. Thus I would love to see a PR landing. 😄 Thanks for your work @georgeraraujo

https://github.com/aardappel/treesheets/pull/641

PR #639, opened 2024-05-02
Also calculate hash while image resampling

tobiolo:


src/mywxtools.h
```diff
@@ -177,3 +177,9 @@ static wxBitmap ConvertBufferToWxBitmap(const vector<uint8_t> &buf, wxBitmapType
     wxBitmap bm(im, 32);
     return bm;
 }
+
+static uint64_t CalculateHash(vector<uint8_t> &idv) {
+    int max = 4096;
+    auto subvector = vector<uint8_t>(idv.begin(), idv.begin() + min(idv.size(), max));
```
aardappel, line 183: This creates a copy.. that is entirely unnecessary. You should be able to pass a `span`, `string_view` or just pointer + size to the hashing function.

PR #643, opened 2024-05-20
Use wxOverlay for hovering on SIMPLERENDER

tobiolo:


tobiolo:
Still during scroll there are some artifacts. @aardappel Do you have any idea? Thanks in advance.

aardappel:
Afraid not too familiar with `wxOverlay`.. seems like it is mostly an improvement for the mac, or?

tobiolo:
Seems to be the case and not a real solution to get same behaviour of Windows on the other platforms.

Issue #644, opened 2024-05-22
/usr/bin/ld: lib/wxWidgets/lib/libwx_gtk3u_core-3.3.a(utilsx11.cpp.o): undefined reference to symbol 'XGetWindowAttributes'


mclaudt:
First of all, many thanks for unbelievably comfort software!

While building under Ubuntu 20.04.3 LTS, ld error appears:

```
/usr/bin/ld: lib/wxWidgets/lib/libwx_gtk3u_core-3.3.a(utilsx11.cpp.o): undefined reference to symbol 'XGetWindowAttributes'
/usr/bin/ld: /lib/x86_64-linux-gnu/libX11.so.6: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
make[2]: *** [CMakeFiles/treesheets.dir/build.make:115: treesheets] Error 1
make[1]: *** [CMakeFiles/Makefile2:449: CMakeFiles/treesheets.dir/all] Error 2
make: *** [Makefile:136: all] Error 2
```

### Steps to reproduce:

Assume we have Linux with git installed, and cloned the repo.

1. `cd treesheets`

2. `cmake -S . -B _build -DCMAKE_BUILD_TYPE=Release`

3. `cmake --build _build`

### Solution:

Add this section to CMakesLists.txt:

```
if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
    find_package(X11 REQUIRED) 
    target_link_libraries(treesheets PRIVATE ${X11_LIBRARIES})  
endif()
```

PR #645, opened 2024-05-22
Configured build to link with X11 libraries on Linux

mclaudt:
Configured build to link with X11 libraries on Linux to fix https://github.com/aardappel/treesheets/issues/644

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -117,6 +117,22 @@ target_link_libraries(
     lobster-impl
 )
 
+######## X11 LINUX BUILD ISSUE ########
+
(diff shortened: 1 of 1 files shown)


aardappel:
This seems like an inoffensive solution, but I am still worried this solves a problem only for a particular distro or install, since noone else has had this particular problem?
@tobiolo

tobiolo:
> This seems like an inoffensive solution, but I am still worried this solves a problem only for a particular distro or install, since noone else has had this particular problem?
> 
> @tobiolo 

Right, I did not experience that problem on my distribution and it does build on the CI so maybe keep this as a distribution-specific patch.

PR #650, opened 2024-05-29
Import and export XML with color attributes in hexadecimal representation

tobiolo:
Due to the autodetection of the representation, `std::stoi` will convert both the decimal and the hexadecimal representation, thus preserving backwards compatibility.


tobiolo:
It will not use a leading # as used in web, but 0x. I think it is still a good compromise because the leading 0x lead to the correct detection, it is still more human readable and using the autodetection enables us to still use the decimal representation.

tobiolo:
Thanks for your review and feedback! ☺️

PR #657, opened 2024-06-01
Adapt XML exporter logic to the simpler XML importer logic

tobiolo:
The XML importer uses `void System::FillXML(Cell*, wxXMLNode*, bool)` that does not make use of inheritance by comparing with the parent cell, while the XML exporter in `wxString Cell::ToText(int, const Selection&, int, Document*, bool)` makes use of inheritance to avoid emitting repeating attribute values.

This commit adapts the XML exporter to the simpler logic of the XML importer.

### DIFF
--- a/src/cell.h
+++ b/src/cell.h
@@ -231,17 +231,17 @@ struct Cell {
                 str.Prepend(wxString() << -text.relsize);
                 str.Prepend(L" relsize=\"");
             }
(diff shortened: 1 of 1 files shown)


aardappel:
Ah, and this also fixes the wrong background colors?

tobiolo:
> Ah, and this also fixes the wrong background colors?

Yes, it seems to do so.

PR #661, opened 2024-06-03
Also import and export user_grid_outer_spacing in XML format and do not emit default values

tobiolo:


src/grid.h
```diff
@@ -622,8 +622,17 @@ struct Grid {
...
         const int grid_border_width =
             cell == doc->rootgrid ? root_grid_spacing : user_grid_outer_spacing - 1;
-        Formatter(r, format, indent,
-                  wxString::Format(L"<grid bordercolor=\"0x%06X\">\n", bordercolor),
+
+        wxString xmlstr(L"<grid");
+        if (bordercolor != 0xA0A0A0) {
+            xmlstr.Append(wxString::Format(wxT(" bordercolor=\"0x%06X\""), bordercolor));
+        }
+        if (user_grid_outer_spacing != 3) {
```
aardappel, line 630: `3` is probably a constant somewhere, and if not it should be?

src/system.h
```diff
@@ -568,6 +570,7 @@ struct System {
                 c->AddGrid(1, numrows);
                 c->grid->bordercolor = std::stoi(
                     n->GetAttribute(L"bordercolor", L"0xA0A0A0").ToStdString(), nullptr, 0);
+                c->grid->user_grid_outer_spacing = wxAtoi(n->GetAttribute(L"outerspacing", L"3"));
```
aardappel, line 573: and here

src/grid.h
tobiolo: Definitely it should be! Thanks for your feedback. Pushed an extra commit for assigning the magic number to a global variable.

tobiolo: Thanks for your review and time!

PR #667, opened 2024-06-08
Offer choice to drop TreeSheets file to open it

tobiolo:


src/mywxtools.h
```diff
@@ -23,7 +23,23 @@ struct DropTarget : wxDropTarget {
         GetData();
         TSCanvas *sw = sys->frame->GetCurTab();
         sw->SelectClick(x, y, false, 0);
-        sw->doc->PasteOrDrop(*sw->doc->dndobjt, *sw->doc->dndobji, *sw->doc->dndobjf);
+        // The data objects in the composite data object for drag and drop operations
+        // are not cleared after the drop operation, leaving the data in the data objects.
```
aardappel, line 27: Having to do all this checking would kinda speak to just constructing these objects anew on each operation? wouldn't code be a lot simpler?

tobiolo: I tried it. Unfortunately, it does not work (does not paste / drop anything) :-(

PR #668, opened 2024-06-08

georgeraraujo:
As done in #281 and #627, also reword question to make it more coherent and cohesive with the opening statement. Update translation template and pt_BR translation.


tobiolo:
Thanks @georgeraraujo!

PR #672, opened 2024-06-13
Manual memory management on image data

tobiolo:
Replace vector for image data with raw array
This allows faster/constant time access to image data

### DIFF
--- a/src/document.h
+++ b/src/document.h
@@ -199,9 +199,8 @@ struct Document {
                 if (image.trefc) {
                     fos.PutC(image.image_type);
                     sos.WriteDouble(image.display_scale);
-                    wxInt64 imagelen(image.image_data.size());
-                    sos.Write64(imagelen);
-                    fos.Write(image.image_data.data(), imagelen);
+                    sos.Write64(wxInt64(image.image_size));
+                    fos.Write(image.image_data, image.image_size);
                     image.savedindex = realindex++;
                 }
             }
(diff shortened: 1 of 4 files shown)


tobiolo:
Put some large 4K images into a TreeSheets and the difference is observable.

aardappel:
Can you point me to where exactly the speedup happens?
Because a `vector` should be completely equivalent to a raw pointer, unless you're sharing these raw pointers or something.
And for most purposes the `vector` is better/safer.

tobiolo:
The speed up happens when the wxBitmap is constructed from the wxImage.
If the wxImage is constructed from an array, it seems to be faster.
image

tobiolo:
Well, it does seem to depend on cache etc. (I did some longer trial runs).
Closing it for now.

tobiolo:
Also tested your proposal, but I did not seem any change in the construction stats, so does not seem to be copied?

aardappel:
Where is `InitFromImage` called from? I don't see it in the code.

It seems like that call takes a `wxImage` which has already been constructed at that point, so I don't see how that function can vary..

There is no way that calling `.data()` and `.size()` on a `vector` is any slower than storing these values raw. A vector is just a convenient wrapper around the buffer you created by hand. The only way a `vector` can be slower is if the code is written to cause it to copy, in which case only that copy needs fixing.

I don't mind using a raw pointer if it is absolutely necessary, but I think only if we know for sure what causes the slowdown.

Worth running on windows with https://github.com/VerySleepy/verysleepy, which may give you better output than `gprof`

tobiolo:
image

tobiolo:
It seems that you always need to go via `wxImage` and this something to live with :-). So I am closing this PR for now.

PR #673, opened 2024-06-15
Use unique_ptr to the wxBitmaps in the icon dropdown

tobiolo:


src/mywxtools.h
```diff
@@ -120,8 +120,7 @@ struct ImagePopup : wxVListBoxComboPopup {
 };
 
 struct ImageDropdown : wxOwnerDrawnComboBox {
-    // FIXME: delete these somewhere
-    Vector<wxBitmap *> bitmaps_display;
+    Vector<unique_ptr<wxBitmap>> bitmaps_display;
```
aardappel, line 123: This uses my `Vector` class, which I am not sure interacts correctly with `unique_ptr`.. maybe use `vector` ?

tobiolo: So far it works perfectly (tested it).

aardappel: I suggested this may be a source of issues.. lets replace it with `vector`.

PR #674, opened 2024-06-15
Move initializer list to declaration block

tobiolo:


src/myframe.h
```diff
@@ -21,9 +21,9 @@ struct MyFrame : wxFrame {
                  // device-independent pixels
     std::vector<std::string> scripts_in_menu;
     bool zenmode;
-    ColorDropdown *celldd = nullptr;
```
aardappel, line 24: Initialization directly in the declaration is superior to doing it in the initializer list when possible, so if anything, better to move other fields to the declaration where possible.

tobiolo:
Is this style ok for you?

src/myframe.h
tobiolo: Thanks for the hint! This saves a lot of lines.

src/document.h
```diff
@@ -6,28 +6,26 @@ struct UndoItem {
     Vector<Selection> path, selpath;
     Selection sel;
     unique_ptr<Cell> clone;
-    size_t estimated_size;
+    size_t estimated_size{ 0 };
```
aardappel, line 9: I would just have done `= 0` like we already have in some places. If there's a reason `{}` is better, I'd put a space ahead of the first `{`.

PR #677, opened 2024-06-15
Make instance_checker a unique_ptr

tobiolo:


src/myapp.h
```diff
@@ -117,14 +122,11 @@ struct MyApp : wxApp {
         }
 
         if (single_instance) {
-            instance_checker = new wxSingleInstanceChecker(
```
aardappel, line 120: you have to be a bit careful with moving the initialization.. like if `single_instance` is false, this init isn't needed to run.

tobiolo: Thanks for your review! I now revised the commit.
There is still a unique_ptr, but the managed object will be set to the wxSingleInstanceChecker object on the heap only if single_instance evaluates to true.
Is this ok?

PR #678, opened 2024-06-15
Make pointer to IPCServer a unique_ptr

tobiolo:


aardappel:
This seems good, but has changes from the other PR in it.

tobiolo:
Thanks for your feedback!
Ok, I did removed the previous commit (had some line overlap).

tobiolo:
Would this be ok now?

PR #679, opened 2024-07-07
Update Lobster to latest release

tobiolo:


tobiolo:
@aardappel 
I do need your help here.
Lobster introduced `FileLoader` and uses a datastructure `string_view_nt`. `main.cpp` uses `script-interface.h` but the lobster headers are not included. How to "import" FileLoader properly?

tobiolo:
image

aardappel:
The best solution probably would be to make Lobster not use the `string_view_nt` type in this case (just `string_view`).. but that is a bit more involved.

A much easier but slightly less clean solution is to just include Lobster's `string_tools.h` in `main.cpp`...

WDYT?

tobiolo:
Hello Wouter
Thanks for your help! Very appreciated :+1: 
I managed to get the source compiled now, but for some reason the lobster treesheets functions like `ts_goto..` are not defined anymore? What would I need to do now? I am little bit lost here.
Thanks in advance for your reply!
Best regards,
Tobias

tobiolo:
image

tobiolo:
As far as I can tell this concerns all functions in `lobster_impl.cpp` (on the lobster side of the interface).

aardappel:
Ah, I'm afraid we changed how namespacing works, from using `_` to `.`.

To make that work:

- Change the registration to `RegisterBuiltin(natreg, "ts", "treesheets", AddTreeSheets);`
- Remove the `ts_` from all functions in `lobster_impl.cpp`
- Change all `ts_` to `ts.` in all .lobster examples we have.

Then also, for any users we have that have their own scripts, advice them to make that last change also :)

tobiolo:
> Ah, I'm afraid we changed how namespacing works, from using `_` to `.`.
> 
> To make that work:
> 
> * Change the registration to `RegisterBuiltin(natreg, "ts", "treesheets", AddTreeSheets);`
> * Remove the `ts_` from all functions in `lobster_impl.cpp`
> * Change all `ts_` to `ts.` in all .lobster examples we have.
> 
> Then also, for any users we have that have their own scripts, advice them to make that last change also :)

Hello Wouter
Thanks for your reply with the nice instructions. I applied them and I can confirm that it works now.
I adapted the Visual studio file as well and now let's see whether this build on the CI :-) I am excited.
Thanks again for your reply giving the background information!
Best regards
Tobias

tobiolo:
It builds and runs successfully.

aardappel:
Looks good!

tobiolo:
> Looks good!

Thanks!

PR #680, opened 2024-07-09
Expose Text filtered on Script Interface for Lobster

tobiolo:


tobiolo:
Not so sure about the boolean data types.
Especially in the use case that it is first checked for filtered and depending on the result the text filtered is set or not.

tobiolo:
This might be a first step to move all the view filter options into scripts...

aardappel:
That is fine, in Lobster, `bool` is an enum which is a subtype of `int`

tobiolo:
> That is fine, in Lobster, `bool` is an enum which is a subtype of `int`

Ah, thanks! 😄

PR #681, opened 2024-07-10
Expose Text lastedit on Script Interface

tobiolo:


tobiolo:
This might serve to sort cells by modification datetime.

PR #684, opened 2024-07-22
Some enhancement for Zen mode

tobiolo:


src/myframe.h
```diff
@@ -695,13 +695,15 @@ struct MyFrame : wxFrame {
...
             tb->Realize();
+            tb->Show(!sys->zenmode);
         }
 
         if (showsbar) {
             wxStatusBar *sb = CreateStatusBar(4);
             sb->SetOwnBackgroundColour(toolbgcol);
             SetStatusBarPane(0);
             SetDPIAwareStatusWidths();
+            sb->Show(!sys->zenmode);
```
aardappel, line 706: Rather than calling this "zenmode" which noone knows what exactly it will do, instead have 2 settings for status and toolbar on/off? That is way clearer and also means you can do one not the other

tobiolo: Thanks for your feedback. I split the zen mode into two persistent options. This also makes the entries in the Options menu obsolete I think.

aardappel:
Ok, this seems to now be mixed with a general refactor that changes when and where cfg is written.. this is not only confusing (please in the future keep refactors separate from a new feature so that it is easy to review), but I think this is also worse.. its best to write the new cfg value the moment it changes, just in case people reboot with TS running, or TS crashes etc.. and then wondering why their setting wasn't saved.

tobiolo:
> Ok, this seems to now be mixed with a general refactor that changes when and where cfg is written.. this is not only confusing (please in the future keep refactors separate from a new feature so that it is easy to review), but I think this is also worse.. its best to write the new cfg value the moment it changes, just in case people reboot with TS running, or TS crashes etc.. and then wondering why their setting wasn't saved.

I can see that point now. Luckily I can just put the latest commit off what I did now.
Would it be ok now?

src/myframe.h
```diff
@@ -498,7 +497,10 @@ struct MyFrame : wxFrame {
                  #else
                  _(L"Toggle &Scaled Presentation View\tF12"));
                  #endif
-        MyAppend(viewmenu, A_ZEN_MODE, _(L"Toggle Zen Mode"));
+        viewmenu->AppendCheckItem(A_SHOWTBAR, _(L"Show Toolbar"));
```
aardappel, line 500: why are these moved here? "view" is more about the actual treesheets data view, these are better off in "options" where they were?

tobiolo: Ok, I moved them to the options menu then.

tobiolo:
Are we ready to go? :-)

tobiolo:
Thanks for your review! Good night!

PR #691, opened 2024-08-06
Add handling of special accelerators that wxGTK cannot handle natively

tobiolo:
Note: This would require adaption of the translation files. But I think this is still better than maintaining one string each for wxGTK and for other platforms.


tobiolo:
Simpler alternative was chosen. Closing this PR.

PR #695, opened 2024-08-10
Replace special accelerators for wxGTK

tobiolo:


tobiolo:
I guess that this is the simpler solution for the Gettext utility to extract the strings.

aardappel:
But do the translation files now need to translate each of these strings twice?

tobiolo:
> But do the translation files now need to translate each of these strings twice?

Yes but given the small amount of additional strings that are not likely to be changed I think this is bearable. Otherwise we also have to change the existing string, too.

PR #701, opened 2024-08-19
Make hover shadow customizable

tobiolo:
The hover shadow comes with a trade-off:
(+) It gives better orientation with lots of text in a TreeSheets
(-) The hover XORs also bitmaps which can be obstructive when there is lot of works with images
So give the user the choice about the hover shadow.


aardappel:
Nice, yeah it can give some bad results :(

tobiolo:
> Nice, yeah it can give some bad results :(

Thanks for the merge!

PR #702, opened 2024-08-19
Write option changes directly to the configuration database on disk

tobiolo:
Instead of waiting until the MyFrame class is destructed, write directly to local configuration database/file when option is changed.
This makes for a more consistent logic likewise used in similiar scenarios.


tobiolo:
rebased now

PR #707, opened 2024-08-25
Updated RU translation based on the template.

qua-rus:


tobiolo:
Thanks for the update!

PR #710, opened 2024-09-07
Use modern STL container method to check existence in map

tobiolo:


tobiolo:
Maybe this should be inlined right away?

PR #712, opened 2024-09-07
Use std::set as container for the tags

tobiolo:


tobiolo:
Keep this a draft because it depends on the other PRs

PR #717, opened 2024-09-14
Use std::vector for colwidths

tobiolo:
Use std::vector as columns are modified (inserted/removed) dynamically.

### DIFF
--- a/src/grid.h
+++ b/src/grid.h
@@ -5,7 +5,7 @@ struct Grid {
     // subcells
     Cell **cells;
     // widths for each column
-    int *colwidths {nullptr};
+    std::vector<int> colwidths {};
     // xsize, ysize
     int xs;
     int ys;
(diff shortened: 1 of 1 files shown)


tobiolo:
This aims to make use of conveniences provided by the STL and thus to increase maintainability of the code

src/grid.h
```diff
@@ -5,7 +5,7 @@ struct Grid {
     // subcells
     Cell **cells;
     // widths for each column
-    int *colwidths {nullptr};
+    std::vector<int> colwidths {};
```
aardappel, line 8: `{}` not needed here.

aardappel (commented):
nice simplification :)

tobiolo:
You are welcome :+1:

src/grid.h
tobiolo: Ah thanks, pushed a commit for this.

PR #719, opened 2024-09-20
Use iterators to travers through strings

tobiolo:
Documentation says at <https://docs.wxwidgets.org/latest/classwx_string.html> that accessing the character by index should be avoided.


aardappel:
And nicer code anyway :)

PR #721, opened 2024-09-22
Let std::format do the work

tobiolo:


src/text.h
```diff
@@ -32,28 +32,7 @@ struct Text {
...
-        if (s[0] == '-') max_significant++;
-        if (s.length() > max_significant) s.erase(max_significant);
-        // Now strip unnecessary trailing zeroes.
-        while (s.back() == '0') s.pop_back();
-        // If there were only zeroes, remove '.'.
-        if (s.back() == '.') s.pop_back();
-
-        t = s;
-    }
+    void SetNum(double d) { t = std::format("{:g}", d); }
```
aardappel, line 35: This is unlikely to produce the same formatting as the code below. The existing code does a lot of specific things to achieve a precise formatting.

tobiolo: But is this stuff really necessary?

tobiolo: Thanks for your explanation! Thus I close this PR.

PR #724, opened 2024-10-28
Use STL binary search algorithms

tobiolo:
This requires itercells to change to std::vector

### DIFF
--- a/src/document.h
+++ b/src/document.h
@@ -84,7 +84,7 @@ struct Document {
     int editfilter {0};
     wxDateTime lastmodificationtime;
     std::set<wxString> tags;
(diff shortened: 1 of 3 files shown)


src/document.h
```diff
@@ -2318,8 +2317,18 @@ struct Document {
         searchfilter = false;
         scrolltoselection = true;
         CollectCells(rootgrid);
-        loopv(i, itercells) itercells[i]->text.filtered =
```
aardappel, line 2321: this particular change now needs and extra `sort`, calls to `lower_bound` and `upper_bound` and generally a lot more complex code.. why?

PR #735, opened 2024-10-30
Use Smart pointers

tobiolo:
Make the pointers to the UndoItem to smart pointers.
Make the unique_ptr to Grid in evaluator to a shared_ptr because this what they are for.

### DIFF
--- a/src/evaluator.h
+++ b/src/evaluator.h
@@ -142,7 +142,7 @@ struct Evaluator {
                     if (g->xs == 1 || g->ys == 1) {
                         return op->runl(g);
                     } else {
-                        std::vector<unique_ptr<Grid>> gs;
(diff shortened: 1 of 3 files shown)


src/evaluator.h
```diff
@@ -142,7 +142,7 @@ struct Evaluator {
                     if (g->xs == 1 || g->ys == 1) {
                         return op->runl(g);
                     } else {
-                        std::vector<unique_ptr<Grid>> gs;
+                        std::vector<shared_ptr<Grid>> gs;
```
aardappel, line 145: I see no reason to use `shared_ptr` here.. they are way more heavyweight than `unique_ptr`, and harder to know what is going to happen.. would like to avoid them unless absolutely needed.

PR #736, opened 2024-10-31
Manage images with unique_ptr

tobiolo:
ThreadPool cannot deal with unique_ptr in a straight-forward way, so leave it out for now.


aardappel:
Wait, can you please explain? I doubt there is any problem with `unique_ptr` and threads, there is likely a different issue.

tobiolo:
The problem is it seems is that I cannot pass the reference to the unique pointer into the lambda function.

tobiolo:
Sorry I could have waited.

aardappel:
Ok, but those errors can probably be solved. I am not sure why that would warrant removing all the multi-threading stuff, which presumably was useful.

In general, you use `unique_ptr` to represent the pointer that OWNS the object, not necessarily for passing pointers. It is fine to pass as `T *` as long as you're sure that the `unique_ptr` outlives it.

tobiolo:
You are perfectly fine. This is why I would like to reintroduce ThreadPool with the other PR. I now found a way to do it and your hint was the one I needed.

PR #737, opened 2024-10-31
Reintroduce Threadpools

tobiolo:


tobiolo:
I finally found a way to use the unique_ptr by making use of the raw pointer.

PR #739, opened 2024-11-01
Declare installer DPI-aware

tobiolo:
See <https://nsis.sourceforge.io/Reference/ManifestDPIAware>


tobiolo:
This creates a super-crisp installer on HiDPI on Windows.

PR #741, opened 2024-11-02
Dark mode on Windows

tobiolo:


tobiolo:
Dark mode is still in a very very bad state on Windows port of wxWidgets. Refrain from it.

PR #743, opened 2024-11-02
Precompiled stdafx in CMake

tobiolo:


tobiolo:
These header guards are necessary otherwise compilation will fail due to redefinition.

tobiolo:
I hope to speed up recompilation within CMake.

PR #745, opened 2024-11-02
Add optional header guards

tobiolo:
This not really necessary right now, but it is a good future-proof measure anyway.

### DIFF
--- a/src/cell.h
+++ b/src/cell.h
@@ -1,3 +1,5 @@
+#ifndef TREESHEETS_CELL_H
+#define TREESHEETS_CELL_H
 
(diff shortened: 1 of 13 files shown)


aardappel:
Not sure why that would be a good idea, as that would indicate they are meant to be possibly included multiple times, which they are not.

TreeSheets uses this system: https://strlen.com/java-style-classes-in-c/

You could try to undo that system, but that likely would require a lot of changes. As long as headers are included inside a struct, you probably would want to do the opposite: error on them being included more than once.

tobiolo:
Thanks for pointing to the article!

PR #746, opened 2024-11-02
Use GraphicsContext with wxDC API

tobiolo:


src/document.h
```diff
@@ -610,9 +610,20 @@ struct Document {
...
 
-    void Draw(wxDC &dc) {
+    void Draw(
+        #ifdef __WXMSW__
+            wxBufferedPaintDC
+        #else
+            wxPaintDC
+        #endif
+            &pdc) {
+        wxGraphicsContext *context = wxGraphicsContext::Create(pdc);
```
aardappel, line 620: What is the advantage of drawing thru this?

Note that we are using XOR rendering on Windows in certain places, which apparently this doesn't support.. what does that look like now?

PR #747, opened 2024-11-03
Make units on the canvas device-independent

tobiolo:
When the coordinates are used for drawing, they are translated to device-dependent units with wxWindow::FromDIP.

### DIFF
--- a/src/cell.h
+++ b/src/cell.h
@@ -156,7 +156,8 @@ struct Cell {
         }
 
         if (drawstyle == DS_GRID && actualcellcolor != parentcolor) {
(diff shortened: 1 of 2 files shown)


aardappel:
What does the before and after look like, visually?

The `FromDIP` everywhere is not great, code wise.

I'd worry this has unintended side effects.. there's a lot more places where sizes are being worked with in very specific ways, including font sizes, rounding sizes etc.

tobiolo:
Well I derived this from the wxWidgets sample files, see e.g. https://github.com/wxWidgets/wxWidgets/blob/master/samples/drawing/drawing.cpp
My hopes are that this may fix printing on the PrintDC. But I still have to validate this. So this PR can be left open.

PR #748, opened 2024-11-03
Make MacOSX Bundle with working translations

tobiolo:
- To create bundle, both build and install must be run by CMake
- Localizations now need to be extra declared in Info.plist


tobiolo:
After adjusting the CI we might also throw away the XCode files.

aardappel:
LGTM

tobiolo:
Thanks for looking into!

PR #749, opened 2024-11-04
Do not use invalid iterators

tobiolo:


tobiolo:
This caused walking and creating the drawpath to malfunction in case of zoom :-( Now fixed.

PR #751, opened 2024-11-05
Avoid redefinition issues with precompiled headers via pragma once

tobiolo:


tobiolo:
This should be more in line with the general approach.

PR #753, opened 2024-11-08
Only allow one of the data types

tobiolo:
On Linux I encountered the bug that a file path was treated both as file path data object and as text object.

### DIFF
--- a/src/document.h
+++ b/src/document.h
@@ -2034,7 +2034,6 @@ struct Document {
     void PasteOrDrop(const wxTextDataObject &pdataobjt, const wxBitmapDataObject &pdataobji,
                      const wxFileDataObject &pdataobjf) {
         wxBusyCursor wait;
-        bool wantsrefresh = false;
 
         if (pdataobjf.GetFilenames().size() != 0) {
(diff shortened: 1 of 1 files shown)


aardappel:
This seems an entirely different change from your description :)

I would guess `wantsrefresh` was used in the past because there was a reason `Refresh` should be called at the end, or to avoid calling it multiple times. Did you verify neither of those reasons are valid anymore, and if so how?

tobiolo:
When I used `wantsrefresh` I had the use case in mind that there could be different text object, file data object and bitmap object at the same time provided from the clipboard for pasting. So I wanted to defer the action of `Refresh()` to the end if e.g. both bitmap and text is pasted. But I now came to the conclusion that this use case is not worth it because when file data object and the text objects is the same in the case of file paths, it gets pasted twice which does not make sense to the user. So I just want to return the function as soon as one data object is hit and pasted. What would you think of it?

tobiolo:
So I checked the branches that toogle the variable and just Refresh() and return then. Is this valid?

aardappel:
I am not too familiar with the Refresh behavior any more :)

PR #756, opened 2024-11-11
Scroll to selection after width change or zoom

tobiolo:


tobiolo:
When the width is changed or in case of zoom, the selection should be always within the visible area.

tobiolo:
In some cases this leads to huge jumps to other edges. Not sure what is better...

aardappel:
Hard to say without seeing what it looks like.

PR #764, opened 2024-11-15
Structure folder for platforms

tobiolo:


aardappel:
can you rebase?

tobiolo:
> can you rebase?

Yes, I did.

aardappel:
I don't see the advantage of putting a `platforms` dir in between, this creates hierarchy where none is needed, given that we have so few dirs anyway. You can instead improve the existing code structure by just renaming the `treesheets` dir to `win64` or better just `win` since `linux` and `osx` don't specify their bit-width either :) We don't have `win32` builds anymore.

aardappel:
yup looks good!

PR #770, opened 2024-11-20
Flatten directory structure

tobiolo:


aardappel:
This mixes licence/author files from 2 unrelated sources?

aardappel:
But now nuvola don't have any docs referring to its origin/license? maybe that should be in each dir that is from nuvola, if you want this structure?

PR #772, opened 2024-11-23
Make rootgrid an unique_ptr

tobiolo:


src/system.h
```diff
@@ -210,7 +210,7 @@ struct System {
 
     void LoadOpRef() { LoadDB(frame->GetDocPath(L"examples/operation-reference.cts")); }
 
-    Cell *&InitDB(int sizex, int sizey = 0) {
+    unique_ptr<Cell> &InitDB(int sizex, int sizey = 0) {
```
aardappel, line 213: This is not great use of `unique_ptr`. `unique_ptr` is supposed to represent the owner. To borrow a `unique_ptr` just use a regular pointer.

A reference to a `unique_ptr` is funny.. it would maybe make sense if you were trying to modify the contents of the `unique_ptr` elsewhere but that doesn't seem to be the case here.

src/document.h
```diff
@@ -107,9 +107,7 @@ struct Document {
         dndobjc->Add(dndobjf);
     }
 
-    ~Document() { DELETEP(rootgrid); }
-
-    uint Background() { return rootgrid ? rootgrid->cellcolor : 0xFFFFFF; }
+    uint Background() { return rootgrid.get() ? rootgrid->cellcolor : 0xFFFFFF; }
```
aardappel, line 110: you can just use `rootgrid` as a condition to test for `null`

src/document.h
```diff
@@ -570,7 +568,7 @@ struct Document {
         ResetFont();
         dc.SetUserScale(1, 1);
         curdrawroot = WalkPath(drawpath);
-        int psb = curdrawroot == rootgrid ? 0 : curdrawroot->MinRelsize();
+        int psb = curdrawroot == rootgrid.get() ? 0 : curdrawroot->MinRelsize();
```
aardappel, line 571: no need for the `.get()`

src/document.h
```diff
@@ -617,7 +615,7 @@ struct Document {
         redrawpending = false;
         dc.SetBackground(wxBrush(wxColor(Background())));
         dc.Clear();
-        if (!rootgrid) return;
+        if (!rootgrid.get()) return;
```
aardappel, line 618: no need

src/document.h
```diff
@@ -2002,7 +2000,7 @@ struct Document {
     }
 
     const wxChar *SearchNext(wxDC &dc, bool focusmatch, bool jump, bool reverse) {
-        if (!rootgrid) return nullptr;  // fix crash when opening new doc
+        if (!rootgrid.get()) return nullptr;  // fix crash when opening new doc
```
aardappel, line 2003: no need

src/grid.h
```diff
@@ -608,7 +608,7 @@ struct Grid {
         const int root_grid_spacing = 2;  // Can't be adjusted in editor, so use a default.
         const int font_size = 14 - indent / 2;
         const int grid_border_width =
-            cell == doc->rootgrid ? root_grid_spacing : user_grid_outer_spacing - 1;
+            cell == doc->rootgrid.get() ? root_grid_spacing : user_grid_outer_spacing - 1;
```
aardappel, line 611: no need

tobiolo:
Thanks for your hints!

tobiolo:
@aardappel

aardappel:
There's still comments above about redundant `.get()` not addressed.

tobiolo:
Ok, I addressed the comments. Is it ok now?

aardappel:
You didn't push.

tobiolo:
> You didn't push.

Ah sorry! Now it should be pushed.

aardappel:
Still don't see my comments above addressed. You marked them resolved but there is no change.

tobiolo:
> Still don't see my comments above addressed. You marked them resolved but there is no change.

I reopened those comments so that you can see why they can't be resolved in my opinion.

tobiolo:
Ah I think it can stay as is. Manual Memory management is fine and better for the readability.

aardappel:
> I reopened those comments so that you can see why they can't be resolved in my opinion.

I don't see any comments from you.

> Manual Memory management is fine and better for the readability.

unique_ptr is often better.. I just had some comments on the use of them.

PR #776, opened 2024-12-02
Make rootgrid an unique_ptr

tobiolo:
This is the second approach. I used the wrong comment type (review comment), so maybe this more cleaner now.

### DIFF
--- a/src/document.h
+++ b/src/document.h
@@ -13,7 +13,7 @@ struct UndoItem {
 
 struct Document {
     TSCanvas *sw {nullptr};
-    Cell *rootgrid {nullptr};
+    unique_ptr<Cell> rootgrid;
     Selection hover;
     Selection selected;
     Selection begindrag;
@@ -91,7 +91,7 @@ struct Document {
         CollectCells(par);      \
         loopv(_i, itercells) for (Cell *c = itercells[_i]; c; c = nullptr)
     #define loopallcells(c)     \
-        CollectCells(rootgrid); \
+        CollectCells(rootgrid.get()); \
         for (Cell *c : itercells)
     #define loopallcellssel(c, rec) \
         CollectCellsSel(rec);     \
@@ -106,8 +106,6 @@ struct Document {
         dndobjc->Add(dndobji);
         dndobjc->Add(dndobjf);
     }
-
-    ~Document() { DELETEP(rootgrid); }
 
     uint Background() { return rootgrid ? rootgrid->cellcolor : 0xFFFFFF; }
 
@@ -127,7 +125,7 @@ struct Document {
     }
 
     void InitWith(Cell *r, const wxString &filename, Cell *ics, int xs, int ys) {
-        rootgrid = r;
+        rootgrid.reset(r);
         InitCellSelect(ics, xs, ys);
         ChangeFileName(filename, false);
     }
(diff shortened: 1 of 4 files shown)


src/document.h
```diff
@@ -570,7 +568,7 @@ struct Document {
         ResetFont();
         dc.SetUserScale(1, 1);
         curdrawroot = WalkPath(drawpath);
-        int psb = curdrawroot == rootgrid ? 0 : curdrawroot->MinRelsize();
+        int psb = curdrawroot == rootgrid.get() ? 0 : curdrawroot->MinRelsize();
```
tobiolo, line 571: We need to call `.get()` because curdrawroot is `Cell *` and rootgrid is `unique_ptr<Cell>`.

aardappel: https://en.cppreference.com/w/cpp/memory/unique_ptr/operator_bool

src/system.h
```diff
@@ -464,13 +464,12 @@ struct System {
...
                     if (!doc.Load(fn)) goto problem;
-                    Cell *&r = InitDB(1);
+                    unique_ptr<Cell> &r = InitDB(1);
                     Cell *c = *r->grid->cells;
                     FillXML(c, doc.GetRoot(), k == A_IMPXMLA);
                     if (!c->HasText() && c->grid) {
                         *r->grid->cells = nullptr;
-                        delete r;
-                        r = c;
+                        r.reset(c);
```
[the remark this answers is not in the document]
aardappel: this seems fine

src/system.h
```diff
@@ -487,7 +486,7 @@ struct System {
 
                     if (as.size()) switch (k) {
                             case A_IMPTXTI: {
-                                Cell *r = InitDB(1);
+                                Cell *r = InitDB(1).get();
```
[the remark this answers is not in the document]
aardappel: this IS borrowing a pointer

src/grid.h
```diff
@@ -608,7 +608,7 @@ struct Grid {
         const int root_grid_spacing = 2;  // Can't be adjusted in editor, so use a default.
         const int font_size = 14 - indent / 2;
         const int grid_border_width =
-            cell == doc->rootgrid ? root_grid_spacing : user_grid_outer_spacing - 1;
+            cell == doc->rootgrid.get() ? root_grid_spacing : user_grid_outer_spacing - 1;
```
aardappel, line 611: See my other comment about using it as boolean

src/document.h
tobiolo: I think the precedence order here is:
`int psb = ((curdrawroot == rootgrid.get()) ? 0 : curdrawroot->MinRelsize());`
Is this correct? If yes, I cannot compare `curdrawroot` and `rootgrid` directly?

src/grid.h
tobiolo: Also here:
I think the precedence order is 
`const int grid_border_width = ((cell == doc->rootgrid.get()) ? root_grid_spacing : user_grid_outer_spacing) - 1)`

src/document.h
aardappel: Ah, did not see. You're right!

src/grid.h
tobiolo: Unfortunately they have not, would make things easier...

tobiolo:
Thanks for the merge!

PR #777, opened 2024-12-08
Display amount of cells selected in status bar

tobiolo:


src/system.h
```diff
@@ -441,6 +441,12 @@ struct System {
         }
     }
 
+    void UpdateAmountStatus(Selection &s) {
+        if (frame->GetStatusBar()) {
+            frame->SetStatusText(wxString::Format(_(L"Amount %d"), s.xs * s.ys), 4);
```
aardappel, line 446: maybe `%d cells` is more informative?

tobiolo: That's right! I updated the PR according to your proposal.

tobiolo:
Thanks!

PR #782, opened 2024-12-10
Use unique_ptr for Grid in Cell

tobiolo:


src/grid.h
```diff
@@ -1045,7 +1043,10 @@ struct Grid {
                 }
             }
             if (rest) {
-                swap_(c->grid, rest->grid);
+                Grid *cg = c->grid.release();
```
aardappel, line 1046: https://en.cppreference.com/w/cpp/memory/unique_ptr/swap

tobiolo: fixed! :-)

PR #785, opened 2024-12-14
Use vector for cells in Grid

tobiolo:


tobiolo:
Rationale for this PR:
- make it more readible
- prerequisite for unique_ptrs in the cells vector

src/grid.h
```diff
@@ -18,7 +18,7 @@ struct Grid {
     bool tinyborder;
     bool folded {false};
 
-    Cell *&C(int x, int y) const {
```
aardappel, line 21: You probably want to make a const and non-const version of this function, such that you don't have to make functions like `Save` and others non-const for no reason

tobiolo: Could you explain to me how I can achieve this? I tried to add a second function, but the compiler only throws errors. I suspect that before this commit, cells is just a pointer to some data structure outside of `Grid`. After this commit, `cells` as a vector lives within the data structure and having a reference to the `Cell*` items, the vector itself is modifiable and cannot be marked as const?

aardappel: By also adding a `Cell *C(int x, int y) const {` (note the absence of `&`).
You then need to check that callers of `C()` correctly receive the return value in a `Cell *&` or `auto &` only when they need to modify and the compiler should select the correct one.
`Save()` for example has no need to modify anything.

tobiolo: Ok, solved this by defining an extra macro because auto deduction does not work in the way I wanted.

src/grid.h
```diff
@@ -58,17 +68,13 @@ struct Grid {
                 for (bool _f = true; _f;)   \
                     for (Cell *&c = g->C(x, y); _f; _f = false)
 
-    Grid(int _xs, int _ys, Cell *_c = nullptr)
-        : xs(_xs), ys(_ys), cell(_c), cells(new Cell *[_xs * _ys]) {
+    Grid(int _xs, int _ys, Cell *_c = nullptr) : xs(_xs), ys(_ys), cell(_c), cells(_xs * _ys) {
         foreachcell(c) c = nullptr;
```
aardappel, line 72: doesn't the default constructor for elements of a `vector<Cell *>` initialize elements to null? If so, don't need this extra loop

tobiolo:
Mmmh, still crashing when pasting cells.

src/grid.h
```diff
@@ -519,24 +531,27 @@ struct Grid {
...
+                    c = nc;
+                } else {
+                    Cell *colcell =
+                        ocells[(nxs ? max(0, min(dx - 1, xs - nxs - 1)) : x) +
+                               (nxs ? y : max(0, min(dy - 1, ys - nys - 1))) * (xs - nxs)];
+                    c = new Cell(cell, colcell);
+                    c->text.relsize = colcell->text.relsize;
+                }
+            } else {
+                c = *oit++;
```
aardappel, line 552: This is new, why was that not needed before?

tobiolo: Ok, applied the old style

tobiolo:
> Mmmh, still crashing when pasting cells.

ok fixed

src/grid.h
tobiolo: You are perfectly right!

PR #786, opened 2024-12-14
Save some space in enumerations

tobiolo:


src/main.cpp
```diff
@@ -37,13 +37,13 @@ static const std::array<uint, 42> celltextcolors = {
     0xFF00FF, 0xFF80FF, 0xFFC0FF, 0xE0C0E0, 0x008080, 0x00FFFF, 0x80FFFF, 0xC0FFFF, 0xC0E0E0,
     0x808000, 0xFFFF00, 0xFFFF80, 0xFFFFC0, 0xE0E0C0,
 };
-enum { CUSTOMCOLORIDX = 0 };
```
aardappel, line 40: None of these enums have a name, thus can't be used as a type, so where exactly do these save storage?
Also do not used unsigned types without reason, they can introduce tricky bugs when mixed with signed types.

tobiolo: Ok, got it! `clang-tidy` complained about them, but I see the side effects of that change, so I am closing this PR.

PR #787, opened 2024-12-14
Only rerender images that are actually needed

tobiolo:


aardappel:
please explain more what this change actually does

PR #788, opened 2024-12-14
Reintroduce autopurge of images

tobiolo:


tobiolo:
Rationale: 
- It does not seem to make much sense to keep images in memory that are currently not used.

aardappel:
This would throw the display version of a bitmap away every second, causing them to be regenerated on the next paint? Why would that be a good idea?

tobiolo:
That is correct!

PR #791, opened 2024-12-19
Remove window restoration correction routine

tobiolo:


aardappel:
Why is this not needed anymore?

tobiolo:
Ah thanks for your questioning! I digged a little bit deeper and found the reason. I would like to exclude some checks for GTK.

tobiolo:
Thanks for your review! +1

PR #794, opened 2024-12-25
Fix crash on open TreeSheets files on MacOS

tobiolo:
When file was loaded directly via the open protocol on MacOS,
`System::FileUsed(const wxString&, Document*)` had been called before
the app event loop was entered. It thus tried to access the uninitialized
file system watcher which led to a crash.

Fix this by contructing the file system watcher right away.

### DIFF
--- a/src/myframe.h
+++ b/src/myframe.h
@@ -5,7 +5,7 @@ struct MyFrame : wxFrame {
     wxTaskBarIcon tbi;
     wxMenu *editmenupopup;
     wxFileHistory filehistory;
-    unique_ptr<wxFileSystemWatcher> watcher {nullptr};
+    unique_ptr<wxFileSystemWatcher> watcher {make_unique<wxFileSystemWatcher>()};
     wxAuiNotebook *nb {nullptr};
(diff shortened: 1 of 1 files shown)


aardappel:
These all seem significant changes that possibly affect multiple platforms, some explanation as to what these changes accomplish would be good.

tobiolo:
Thanks for your feedback. I refined the commit message. Would this be ok for you now?

aardappel:
Where is the commit message? Don't see it here on the PR.

Would be better to comment on the individual changes.

tobiolo:
I meant:

image

image

tobiolo:
@aardappel

aardappel:
Ok, I see the file system watcher being constructed immediately, but there are other changes in this PR, which will also affect other platforms.

If the changes are unrelated, then its best to stick them into separate PRs, so those can have their own explanation.

If they are related, I'd like to know why they are needed in addition to moving the FS watcher, and how they don't matter on non-mac platforms.

tobiolo:
Thanks @aardappel for your feedback. I reduced the changes.

PR #797, opened 2025-01-08
const char instead of const wchar_t

tobiolo:
wxWidgets macro now expects `const char*` instead of `const wchar_t*`

```
.../Development/treesheets/src/myframe.h:130:26: error: no matching function for call to 'wxUnderscoreWrapper'
  130 |             wxMessageBox(_(L"Error loading core data file (TreeSheets not installed correctly?)"),
      |                          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
.../Development/treesheets/lib/wxWidgets/include/wx/translation.h:48:48: note: expanded from macro '_'
   48 |     #define _(s)                               wxUnderscoreWrapper((s))
      |                                                ^~~~~~~~~~~~~~~~~~~
...//Development/treesheets/lib/wxWidgets/include/wx/translation.h:347:17: note: candidate template ignored: could not match 'const char' against 'const wchar_t'
  347 | const wxString& wxUnderscoreWrapper(const char (&msg)[N])
      |                 ^
```

### DIFF
--- a/src/document.h
+++ b/src/document.h
@@ -140,7 +140,7 @@ struct Document {
     }
 
     const wxChar *SaveDB(bool *success, bool istempfile = false, int page = -1) {
-        if (filename.empty()) return _(L"Save cancelled.");
+        if (filename.empty()) return _("Save cancelled.");
         Cell *ocs = selected.GetFirst();
         auto start_saving_time = wxGetLocalTimeMillis();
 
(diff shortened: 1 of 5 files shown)


aardappel:
How did this happen? Did we upgrade to a newer wxWidgets? Does that mean it won't compile with older versions after this change?

tobiolo:
Yes, wxWidgets introduced these changes. I cannot say whether it compiles with older versions because when I want to checkout the wxWidgets submodule to v3.2.6 I get issues with a scintilla submodule within wxWidgets.

aardappel:
ok, that be good to figure out first. If it means we're locked to the latest wxWidgets we should be careful about that?

tobiolo:
This will be fixed upstream
https://github.com/wxWidgets/wxWidgets/pull/25063

PR #799, opened 2025-01-09
Use const char* instead of const w_char_t*

tobiolo:
Rationale: wxWidgets macro for localization only uses const char, so let's clean up the whole code. Only use w_char_t where it is absolutely necessary (Windows API).

### DIFF
--- a/src/cell.h
+++ b/src/cell.h
@@ -218,84 +218,84 @@ struct Cell {
         if ((format == A_EXPHTMLT || format == A_EXPHTMLTI) &&
             ((text.stylebits & STYLE_UNDERLINE) || (text.stylebits & STYLE_STRIKETHRU)) &&
             this != doc->curdrawroot && !str.IsEmpty()) {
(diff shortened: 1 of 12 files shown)


tobiolo:
It even makes for a nicer code. We could also instead use std::string with the s suffix but this looks not so cluttered.

tobiolo:
I think as long as we stick to UTF-8 encoding this should work?

aardappel:
See my comments on your other similar PR. While this seems nice (I much prefer UTF-8 everywhere), I don't understand why this is suddenly possible/necessary, and how it affects builds against various versions of wxWidgets.

tobiolo:
You are absolutely right: I asked upstream.
https://github.com/wxWidgets/wxWidgets/commit/3dddaeb8f2bb8fcfc07d44331a8a2224e24db34a#commitcomment-151169404

tobiolo:
This will be fixed upstream
https://github.com/wxWidgets/wxWidgets/pull/25063

PR #808, opened 2025-02-03
Improve grammar when there is just one cell/char

georgeraraujo:


aardappel:
This is more correct, but pretty much no document contains 1 cell or 1 character. So no one is every going to see the improved message.

georgeraraujo:
> This is more correct, but pretty much no document contains 1 cell or 1 character. So no one is every going to see the improved message.

Yes, for the first change is the PR that makes sense. The second change, on the other hand, applies to code that updates the status bar every time the user clicks a cell or a grid line, or selects a number of cells.

georgeraraujo:
Done.

PR #810, opened 2025-02-03
Key bindings for Layout & Render Style menu...

### LINKED ISSUE #809: Key bindings for Layout & Render Style menu don't make it into `TS/translations/ts.pot` because of `CTRLORALT` macro introduced in 5d5ee4a50ecb0f2e10d9d72a4b5c821422f68a0a
The following screenshot shows it:

Image

I'll open a PR with a fix.

georgeraraujo:
... don't make it into `TS/translations/ts.pot` because of `CTRLORALT` macro introduced in 5d5ee4a50ecb0f2e10d9d72a4b5c821422f68a0a , so rearrange code in `src/myframe.h` to bring them back.

Closes #809.

### DIFF
--- a/src/myframe.h
+++ b/src/myframe.h
@@ -386,33 +386,50 @@ struct MyFrame : wxFrame {
             MyAppend(navmenu, A_BROWSEF, _(L"Open &file\tF4"),
                      _(L"Opens up the text from the selected cell in default application for the file type"));
 
-            #ifdef __WXMAC__
-                #define CTRLORALT "CTRL"
-            #else
-                #define CTRLORALT "ALT"
-            #endif
-
             wxMenu *laymenu = new wxMenu();
-            MyAppend(laymenu, A_V_GS,
-                     _(L"Vertical Layout with Grid Style Rendering\t" CTRLORALT "+1"));
-            MyAppend(laymenu, A_V_BS,
-                     _(L"Vertical Layout with Bubble Style Rendering\t" CTRLORALT "+2"));
-            MyAppend(laymenu, A_V_LS,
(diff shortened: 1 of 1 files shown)


aardappel:
I am not a fan of how much duplication this fix causes, there has to be other ways to make this string translatable without copying everything twice.

Like maybe `_(L"Vertical Layout with Grid Style Rendering" "\t" CTRLORALT "+1")` so now the string to translate is just the first part?

Or use actual string concatenation.
`_(L"Vertical Layout with Grid Style Rendering") + (L"\t" CTRLORALT L"+1")` (may need a `wxString` constructor)

georgeraraujo:
I don't like it too. I only did it like that because the `CTRLORALT` macro only gets resolved to `CTRL` or `ALT` at compile time. I read the [documentation](https://www.gnu.org/software/gettext/manual/html_node/xgettext-Invocation.html), and as far as I could tell `xgettext` does not handle this case.

aardappel:
@tobiolo

tobiolo:
> I am not a fan of how much duplication this fix causes, there has to be other ways to make this string translatable without copying everything twice.
> 

I agree. The costs of the duplicates are too high.

> 
> Like maybe `_(L"Vertical Layout with Grid Style Rendering" "\t" CTRLORALT "+1")` so now the string to translate is just the first part?
> 
> 
> 
> Or use actual string concatenation.
> 
> `_(L"Vertical Layout with Grid Style Rendering") + (L"\t" CTRLORALT L"+1")` (may need a `wxString` constructor)

I would prefer this one. I am unsure whether the first one works but I have not yet tried.

PR #816, opened 2025-03-08
Automatic package generation

tobiolo:


georgeraraujo:
Up to the [release just before this commit](https://github.com/aardappel/treesheets/releases/tag/13752012278), the `build-windows` CI action built an installer that didn't require admin rights to run, namely `windows_treesheets_setup.exe` - convenient for regular non-admin users on corporate environments. Would you consider bringing it back?

`windows_treesheets_no_installer.zip` was also useful to have.

tobiolo:
I am working on adding an ZIP Generator to CPack. This should bring back the no installer ZIP.

tobiolo:
CPack ZIP generator is implemented according to
https://cmake.org/cmake/help/latest/cpack_gen/archive.html

I pushed a commit.
ZIP should be listed in Release section as soon as Github Action produced it.

PR #818, opened 2025-03-14
Allow non-privileged installation on Windows

georgeraraujo:
This is useful for environments where users usually do not have elevated rights, like in business environments.


tobiolo:
Thanks @georgeraraujo! Smart move!

PR #821, opened 2025-03-19
Try to get full path on MacOS

tobiolo:
When the executable is launched in its sitting directory, argv[0] is only "./TreeSheets". This will fail MyFrame::GetDataDir because it needs the full path.

See dyld(3) for details with regards to _NSGetExecutablePath.


tobiolo:
readlink somehow gives the relative path on macOS.

PR #828, opened 2025-03-21
Changes to Russian localization based on updated translation template.

qua-rus:


tobiolo:
Hi Alexander,
Thanks for the update!
Kind regards,
Tobias

PR #837, opened 2025-03-24
Set Debian package epoch to 1 to avoid downgrade

tobiolo:


qua-rus:
@tobolio, now the new version of the package is determined correctly. Previously I specified the version for installation and fixed it - after installation.
Thank you.

tobiolo:
You are welcome!

PR #841, opened 2025-03-25
Default colwidth

tobiolo:


aardappel:
Cool.. where is `defaultmaxcolwidth` actually used though?

tobiolo:
> Cool.. where is `defaultmaxcolwidth` actually used though?

- When you create a new document, it is used for the column width.
- When you are in a Document, on the highest level when you create a new column.

aardappel:
Ah sorry, the variable is already used, that wasn't obvious from the commit

PR #845, opened 2025-03-27

### LINKED ISSUE #842: The ts.pot file lacks lines for translating...
The ts.pot file lacks lines for translating the Menu item:` Options / Set default column width...`:

Image

There are also no lines for translating lines from the called window:

Image

georgeraraujo:
Closes #842


tobiolo:
Thanks!

PR #859, opened 2025-04-12
Add single steps in scripts to undo stack

tobiolo:
This has following trade-off:

Positive:
+ Only marks documents as modified when they are (undo stack is used)
+ Can undo single steps in scripts which is good when one step does not do what it was intended to do

Negative:
- Script modifications cannot be undone all at once.


aardappel:
The other negative is that any script that makes thousands of changes now adds thousands of undo items, which may really slow it down.

Then again, I guess most scripts are either exporters (which shouldn't create any undos) or scripts that make a reasonable amount of modifications, so probably ok?

PR #861, opened 2025-04-12
Only add to undo stack once for modifications

tobiolo:


aardappel:
Why do you prefer this approach now? Do you want to be to undo all changes of a script at once?

tobiolo:
Yes, I had second thought about it and I think it is a more optimal solution so that we use the undo stack once for the script block but only when we are about to change the Document.

aardappel:
Ok, fair enough.

tobiolo:
Thanks for having had a look into it!

PR #862, opened 2025-04-12
Provide artifacts on Pull Requests

tobiolo:


tobiolo:
To try and test out binaries on Windows and Mac.
Linux is not supported because the `upload-artifact` action does not support colons in filenames that the Debian CPack generator produces because of the epoch.

aardappel:
What is this for? So people can test PRs without building?

tobiolo:
Exactly.

PR #863, opened 2025-04-13
Manage Lobster scripts via separate wxFileHistory object and wxConfig group

tobiolo:


tobiolo:
Rationale: the scripts folder lies within the Application bundle on MacOS and is thus not easily accessible for the user. Also on Linux, the scripts directory is at /usr/... that is not writable by default for users. So encourage the user to add own scripts that can be placed anywhere.

aardappel:
Good idea :)

Does the file history have a max number of entries? Because we add the default ones to it.

aardappel:
Ah, max a 1000, that should be sufficient :)

tobiolo:
Exactly, I do hope so, too!

tobiolo:
Thanks for having taken a glance on it!

PR #865, opened 2025-04-14
Fix NSIS

tobiolo:


tobiolo:
We are not completed at the finish line, but these are the first steps. Now we have only the project name as folder directory and package name.

tobiolo:
https://nsis.sourceforge.io/Reference/SetShellVarContext

https://nsis.sourceforge.io/Docs/Chapter4.html

> $APPDATA

> The (roaming) application data directory. The context of this constant (All Users or Current user) depends on the [SetShellVarContext](https://nsis.sourceforge.io/Docs/Chapter4.html#setshellvarcontext) setting. The default is the current user.

> Detection of the current user path requires Internet Explorer 4 and above. Detection of the all users path requires Internet Explorer 5 and above.

> This constant is not available on Windows 95 unless Internet Explorer 4 with Active Desktop is installed.

> $LOCALAPPDATA

> The local (non-roaming) application data directory. The context of this constant (All Users or Current user) depends on the [SetShellVarContext](https://nsis.sourceforge.io/Docs/Chapter4.html#setshellvarcontext) setting. The default is the current user. The All Users location is also known as %ProgramData% on Vista and above.

> This constant is available on Windows ME, 2000 and above.

tobiolo:
I think they messed up the ShellVarContext.

aardappel:
It needs to not be installed in a "roaming" location though? And it should not be the "data" directory. Isn't that what got it in the onedrive folder?

PR #867, opened 2025-04-21
Provide Linux artifact on PRs

jn64:
Follow up to https://github.com/aardappel/treesheets/pull/862#issuecomment-2799036529


tobiolo:
Thanks @jn64!

PR #868, opened 2025-04-22
Add dependabot.yml to auto-PR submodule updates on a weekly basis

tobiolo:


tobiolo:
Some third-parties like Flatpak cannot be configured to use `--remote` for git submodules to update them from the remote. Thus the submodule commit pointer has to be updated to a recent commit.

PR #871, opened 2025-04-22
Workflow condition for push vs PR

jn64:
Use the event name which is more explicit and easier to understand
when certain steps are supposed to run.

Event names are same as what is given to `on:` at the start of the file,
e.g. `push`, `pull_request`


tobiolo:
Much cleaner and more readible now! Thanks @jn64! :-)

PR #875, opened 2025-04-25
Enable Windows ARM builds

tobiolo:


tobiolo:
This requires tinycc being properly updated. Unfortunately using latest tinycc with Windows ARM support crashes TreeSheets on macOS executing Lobster scripts. 😮‍💨

aardappel:
* You can't just update libtcc here, if it is not updated in Lobster, it will get overwritten next time we update Lobster.
* Don't add new files for things we don't use, like the coff format etc.
* If this version crashes Mac we likely can't merge it? Where does it crash, exactly?

tobiolo:
You are right. I will wait until Windows ARM may be supported properly sometime by Lobster.

PR #881, opened 2025-04-26
Use Windows API more directly

tobiolo:


aardappel:
This may stop our code working on older versions of Windows, but I guess that is ok by now..

PR #890, opened 2025-05-01
Provide infrastructure for ImGUI integration

tobiolo:
This is intended for discussion.


src/im/main.cpp
```diff
@@ -0,0 +1,208 @@
+// Dear ImGui: standalone example application for GLFW + OpenGL 3, using programmable pipeline
```
tobiolo, line 1: This is just the demo to demonstrate that the build infrastructure works.

CMakeLists.txt
```diff
@@ -155,6 +155,29 @@ endif()
...
+find_package(OpenGL REQUIRED)
+
+add_executable(
+    trimsheets
+    src/im/main.cpp
+    thirdparty/imgui/imgui_demo.cpp
+    thirdparty/imgui/imgui_draw.cpp
+    thirdparty/imgui/imgui_tables.cpp
+    thirdparty/imgui/imgui_widgets.cpp
+    thirdparty/imgui/backends/imgui_impl_glfw.cpp
```
tobiolo, line 170: @aardappel 
What would you think which backend should fit for cross-platform use the best? Have you any hints.
I would appreciate this greatly. TIA!

aardappel: Lobster (and my game engine) uses imgui extensively, and there we use the opengl3 + sdl2 backends..

aardappel:
Can you please explain a bit more about what you're thinking of doing?

I am very familiar with Dear ImGui, and I love it, but I would think that trying to replace all our wxWidgets functionality with imgui equivalents would be a bit crazy undertaking. They are so very different. And wxWidgets contains a lot of functionality that imgui doesn't have.

Trying to separate it would be hard. You don't want 2 copies of all our treesheets code, as that would be a maintenance nightmare, but trying to pull out only the wxWidgets functionality cleanly would be super hard too.

imgui is also very much a "programmer style" ui with very few styling/theming options, so you'd go from something that looks kinda like a native app on most platforms to something that looks like a debug video game UI :)

It also doesn't help with porting this to new platforms that wxWidgets doesn't reach, like mobile.. because mobile would need a different kind of UI to work on small screens that is exactly what imgui is not good at.

The only advantage I see is that the actually cell/grid/text rendering could be moved to OpenGL, and would look a lot better, maybe allow for arbitrary scaling and be a lot faster.

tobiolo:
Thanks for your feedback on this idea. It led me also come to the conclusion that we should better refrain from elaborating more on the idea of porting TreeSheets to ImGUI. But maybe we can leverage `wxGLCanvas` (see https://wiki.wxwidgets.org/WxGLCanvas for examples). Do you think that is feasible and worthwile?

aardappel:
Hard to say, there may be platform issues and likely 101 corner cases to resolve. You could give it a go to see how much work it is.

tobiolo:
Well, I might give it a try but this does not appear to be a high priority issue. There is still other stuff that is annoying me highly on macOS. For some reason I cannot explain on macOS the window does not scroll to the initial selection when I open a TreeSheets file. For some reason I cannot explain I have to move the mouse etc. so that it refreshes and then it is ok. Also the screenshots on macOS cannot be pasted into TreeSheets because it seems to be of PNG file format instead of a Bitmap, as far as I can judge.

PR #918, opened 2025-05-13
Prefer wxString over wxChar

tobiolo:


src/document.h
```diff
@@ -521,14 +521,14 @@ struct Document {
         DrawSelectMove(dc, selected, true, false);
     }
 
-    const wxChar *NoSel() { return _(L"This operation requires a selection."); }
-    const wxChar *OneCell() { return _(L"This operation works on a single selected cell only."); }
-    const wxChar *NoThin() { return _(L"This operation doesn't work on thin selections."); }
-    const wxChar *NoGrid() { return _(L"This operation requires a cell that contains a grid."); }
+    const wxString &NoSel() { return _(L"This operation requires a selection."); }
```
aardappel, line 524: How does this work? This forces a string object construction, and then you return a reference to it, which would create a dangling reference.

You either need to make the return type just `wxString` so it can move/copy the object (hopefully RVO).

But really, why all that allocation? Do we have a case where we want to return a string that is not a constant?

tobiolo: `_()` translates to `wxGetTranslation` `https://docs.wxwidgets.org/latest/group__group__funcmacro__string.html#ga49dcdc17d82a627b7a50d401f1a37d69 and it returns the `wxString` by reference. So there does not seem to be further allocations.

tobiolo: But you are completely right, string literal L"" means wide character pointer so this is very redundant. I reverted it. Thanks for your feedback.

aardappel: The docs for that say `If the string is not found in any of the loaded message catalogs the original string is returned`. The original string is also a `const wxString &` argument, which refers to the stack allocated conversion of the literal string (which will do allocations for larger strings).

So unless `wxGetTranslation` stores a copy and always refers to that, the original code before your change possibly already was a problem

aardappel: Ok, I was just able to verify that `wxGetTranslation` indeed makes a copy even if the original string is passed thru. So either `const wxString &` or `wxChar *` should work equally well as return value, and `const wxString &` may well be nicer, assuming we can rely on the implementation of `_(` never changes to simply pass thru the literal or temp string constructed from it.

PR #947, opened 2025-06-04
Remove pointer in auto declaration

tobiolo:


aardappel:
Interesting.. I've myself always preferred to use `auto` instead of `auto *` (only using `auto &` since that is necessary), so this is likely more consistent with existing code.

The only advantage of `auto *` is that if the caller changes `*` to `&` you'll get an error.. but I don't think that is that important. But something to be aware of if you ever make something a reference.

PR #956, opened 2025-06-05
Changes to Russian localization (v2.2) based on updated translation t…

qua-rus:
…emplate.


tobiolo:
Thanks for the update!

PR #967, opened 2025-06-06
Global search

tobiolo:
PoC for global search through all Documents


tobiolo:
- Glitches after tab changes on Linux (X11): Wrong selection positioning etc.
How to properly change the Device context after tab change?

tobiolo:
Thanks for your feedback! I will defer this feature because maybe the use case is not as large as I thought it be.

PR #977, opened 2025-06-11
pt_BR: tweak translation and update tutorial

georgeraraujo:


aardappel:
I don't think we need to have translation updates so frequently? Probably once every few months is sufficient?

georgeraraujo:
> I don't think we need to have translation updates so frequently? Probably once every few months is sufficient?

While I see your point, I titled this PR as a "tweak" because it is more about marking strings for translation and updating the pt_BR tutorial than about updating the pt_BR translation. Anyway, I'm fine with you closing it.

PR #985, opened 2025-06-14
WIP: Use less wxClientDC

tobiolo:
- Dissolve DrawSelectMove because only scrolltoselection needs to be set and Refresh triggered
- Use ScrollIfSelectionOutOfView only with wxPaintDC

### DIFF
--- a/src/document.h
+++ b/src/document.h
@@ -230,16 +230,7 @@ struct Document {
         sel.g->DrawSelect(this, dc, sel, cursoronly);
     }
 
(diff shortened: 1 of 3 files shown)


tobiolo:
SearchNext does not work

tobiolo:
Thanks for your feedback!

I agree with your proposal and I will create a separate branch for reviewing the smaller commits and invite you for review. Thanks for the close collaboration on this!

PR #998, opened 2025-06-15
wxPaintDC rewrite

tobiolo:
This Pull Requests serves for discussion of a rewrite.

Background: wxWidgets has deprecated wxClientDC starting with 3.3.0. This is a refactoring to just do the drawing specifics only within the Drawing routine with wxPaintDC.

### DIFF
--- a/src/document.h
+++ b/src/document.h
@@ -32,7 +32,6 @@ struct Document {
     int fgutter {6};
     int lasttextsize;
     int laststylebits;
-    int initialzoomlevel {0};
     Cell *curdrawroot;  // for use during Render() calls
     vector<unique_ptr<UndoItem>> undolist;
     vector<unique_ptr<UndoItem>> redolist;
@@ -550,10 +549,6 @@ struct Document {
         dc.Clear();
         if (!rootgrid) return;
         sw->GetClientSize(&maxx, &maxy);
-        if (initialzoomlevel) {
-            ZoomSetDrawPath(initialzoomlevel);
-            initialzoomlevel = 0;
-        }
         Layout(dc);
         double xscale = maxx / (double)layoutxs;
         double yscale = maxy / (double)layoutys;
--- a/src/document.h
+++ b/src/document.h
@@ -218,16 +218,11 @@ struct Document {
         return _(L"");
     }
 
-    void DrawSelect(wxDC &dc, Selection &sel, bool refreshinstead = false,
-                    bool cursoronly = false) {
+    void DrawSelect(wxDC &dc, Selection &sel) {
         sys->UpdateAmountStatus(sel);
-        if (refreshinstead) {
-            sw->Refresh();
-            return;
-        }
         if (!sel.g) return;
         ResetFont();
-        sel.g->DrawSelect(this, dc, sel, cursoronly);
+        sel.g->DrawSelect(this, dc, sel);
     }
 
     void RefreshMove(Selection &sel) {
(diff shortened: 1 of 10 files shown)


tobiolo:
image

tobiolo:
I have made a list in order to list the callers and callees of functions that use `wxDC`.

aardappel:
> Main idea:
> 
> 1. User requests Action from Frame (OnMenu) or Canvas (OnChar)
> 2. Check whether Action requires drawing or not
> 3. If Action requires drawing, write action number into variable that Canvas OnPaint can evaluate

I don't understand this. If the action requires drawing, it should its non-drawing manipulations, call `Refresh`, and set any state that indicates what needs to change in drawing, if necessary.

> 4. Request drawing with Canvas refresh
> 5. Canvas::OnPaint procedure starts
> 6. in the procedure there are functions before Layout, Render, Select and its end where procedures for the Action number are placed if required

Same here, this should only do things related to drawing (such as shift to center, draw selection etc), not run the whole action.

> 7. Reset Action number variable

Reset any state used by 3 and 6.

> 
> Notes:
> 
> * Document shall not contain any drawing related code und vice versa for Canvas
> * only one draw per action

tobiolo:
> > Main idea:
> > 
> > 1. User requests Action from Frame (OnMenu) or Canvas (OnChar)
> > 2. Check whether Action requires drawing or not
> > 3. If Action requires drawing, write action number into variable that Canvas OnPaint can evaluate
> 
> I don't understand this. If the action requires drawing, it should its non-drawing manipulations, call `Refresh`, and set any state that indicates what needs to change in drawing, if necessary.
> 
> > 4. Request drawing with Canvas refresh
> > 5. Canvas::OnPaint procedure starts
> > 6. in the procedure there are functions before Layout, Render, Select and its end where procedures for the Action number are placed if required
> 
> Same here, this should only do things related to drawing (such as shift to center, draw selection etc), not run the whole action.
> 
> > 7. Reset Action number variable
> 
> Reset any state used by 3 and 6.
> 
> > Notes:
> > 
> > * Document shall not contain any drawing related code und vice versa for Canvas
> > * only one draw per action

Thanks for your comments, good to have the right general idea first.

tobiolo:
Approach is to walk down from the event processing functions that used to create wxClientDC to get rid of wxDC until we hit the hard ground that require drawing to queue them up

tobiolo:
image

This problem is still to be addressed,
Refresh queues operation up,

tobiolo:
@aardappel 
Would you mind to take a look on this?

tobiolo:
So, I have think this is ready now.

sins921 (approved):

src/document.h
```diff
@@ -76,9 +77,14 @@ struct Document {
     wxPrintData printData;
     wxPageSetupDialogData pageSetupData;
     uint printscale {0};
-    bool redrawpending {false};
     bool scrolltoselection {true};
     bool scaledviewingmode {false};
+    bool updatehover {false};
```
aardappel, line 82: would maybe be good to give all these variables the same prefix, so that it is clear these are all used for passing events to Draw

aardappel (commented):
Sorry, was away for 2 weeks travelling!

Looks good! A nice cleanup I think!

tobiolo:
> Sorry, was away for 2 weeks travelling!
> 
> Looks good! A nice cleanup I think!

Thanks! I hope you had a good travel and nice vacations!

PR #1000, opened 2025-06-16
update Korean...

VenusGirl:
Korean update...


VenusGirl (commented):
Change File Name
- ko.po -> ts.po
- ko.mo -> ts.mo

PR #1010, opened 2025-06-18
Update RU translation

qua-rus:
Based on #1004 updated translation template.


tobiolo:
Yes, there should be a update in the status bar at least about what happened after the user had call the menu item action.

PR #1011, opened 2025-06-18
Remove wxClientDC from OnSearch

tobiolo:
Also remove SetSearchTextBoxBackgroundColour because it requires SearchNext that requires wxDC


tobiolo:
I want to reduce complexity here.

PR #1015, opened 2025-06-19
Do not draw hover

tobiolo:
Get rid of implementations that are not supported by all platforms


aardappel:
Why? I'd rather that either it still looks better on the platform that supports it, or that a generally better way to implement the same cross-platform be found.

PR #1019, opened 2025-06-19
Only support simple rendering

tobiolo:
- Client server architecture for device context will not be supported any longer
- This means:
 - Just use SIMPLERENDER because everythings needs to be completely drawn
 - Client color is handled differently on Windows so no custom cursor color there


aardappel:
`SIMPLERENDER` for things like hover over doesn't look as good though, and it seemed these rendering operations still work on Windows? I am not following why this had to be removed.

PR #1022, opened 2025-06-20
Reconstruct toolbar on DPI change

tobiolo:


tobiolo:
This is a clean way to ensure that sizes are correct after DPI changes.

tobiolo:
No, this causes crashes.

PR #1032, opened 2025-07-03
Updated POT template file. Update RU translation. …

qua-rus:
… Added template update command file - pot_update.sh. Existing data clarified.


qua-rus:
I hope I did everything right.
There are warnings:
```
$ ./pot_update.sh
xgettext: Option '--sort-output' is deprecated.
document.h:223: warning: Empty msgid. It is reserved for GNU gettext:
                         gettext("") returns the header text with meta information,
                         not the empty string.
document.h:1243: warning: Empty msgid. It is reserved for GNU gettext:
                         gettext("") returns the header text with meta information,
                         not the empty string.
```
xgettext [Output details](https://www.gnu.org/software/gettext/manual/html_node/xgettext-Invocation.html#Output-details)

tobiolo:
Thanks for your work!

PR #1040, opened 2025-07-09
Use FetchContent for wxWidgets dependency

tobiolo:


tobiolo:
Thanks for this feedback! :) Highly appreciated!

PR #1042, opened 2025-07-11
Externalize Lobster in CMake

tobiolo:


aardappel:
Nice :)

tobiolo:
Yeah, that was fun! :)

PR #1054, opened 2025-07-14
Do not name unused function parameters

tobiolo:


src/lobster_impl.cpp
```diff
@@ -12,175 +12,177 @@ namespace script {
...
-nfr("set_window_size", "width,height", "II", "", "resizes the window",
-    [](StackPtr &sp, VM &vm, Value w, Value h) {
-        si->SetWindowSize(w.intval(), h.intval());
-        return NoVal();
-    });
-
-nfr("last_edit", "", "", "I",
-    "gets the timestamp of the last edit in milliseconds since the Unix/C epoch",
-    [](StackPtr &sp, VM &vm) { return Value(si->LastEdit()); });
+    nfr("goto_root", "", "", "",
```
aardappel, line 15: To not indent these was actually intentional.. they have to be inside a function, but the indentation here is not helpful when it is such a long list of declarations.

PR #1061, opened 2025-07-17
Only remember open files on close

tobiolo:


tobiolo:
Only doing the action once on closing the app I think is sufficient.

aardappel:
Ok, but why? What's the downside on doing it when it changes?

My original reason for doing that is that sometimes TS crashes, or more mundanely, it is killed as part of a shutdown etc.. when you had it running for a long time, its annoying to have setting changes you made long ago be forgotten.

PR #1071, opened 2025-07-24
Implement Statistics in Lobster

tobiolo:
This is now deemed as a non-essential function that can be implemented in Lobster.


src/document.h
```diff
@@ -1228,15 +1228,6 @@ struct Document {
             case A_FILTERS: SetSearchFilter(true); return nullptr;
             case A_FILTEROFF: SetSearchFilter(false); return nullptr;
 
-            case A_SHOWSTATS: {
-                int numcells = 0, textbytes = 0;
-                selected.g->GetStats(selected, numcells, textbytes);
```
aardappel, line 1233: seems there's still a `GetStats` that could be removed.

PR #1118, opened 2025-08-12
Express intentions of variables more explicit

tobiolo:


aardappel:
Why suddenly the replacement of `auto` ? We use `auto` extensively, and these don't seem particularly confusing cases of it.

tobiolo:
You are right, I am going to correct this.

PR #1137, opened 2025-08-18
Polymorphism does not work for wxDataObjectSimple on Windows

tobiolo:


aardappel:
What a mess.. we've made so many changes to this code in the past, wxWidgets seems to really be quite bad here!

PR #1157, opened 2025-08-30
Script can operate on the complete document

tobiolo:


src/treesheets_impl.h
```diff
@@ -14,8 +14,8 @@ struct TreeSheetsScriptImpl : public ScriptInterface {
 
     void MarkDocAsModified() {
         if (docmodified) return;
-        // Consider the case that script operates on multiple cells in a grid
-        doc->AddUndo(cur->parent ? cur->parent : cur);
+        // The script can operate on multiple cells throughout the document
+        doc->AddUndo(doc->root);
```
aardappel, line 18: This is worse.. `MarkDocAsModified` is called for each script operation. This would copy the whole document into the undo for each one of them. What are you trying to solve, exactly?

aardappel: Ah ok, its only done once, missed that.

But if it moves to the parent cell, if it modifies anything there, it would create an undo for that cell? I still don't see why we need to save an undo for the whole doc, or even the parent.

It depends a bit on the situation, if a script modifies pretty much the whole doc, then this is more efficient. But for scripts that only touch the current selection, adding undo for the whole doc seems wasteful.

tobiolo: We could separate marking the document as modified and adding to the undo stack?

tobiolo: And there is also the case then we we add each cell to the undo stack and we have an script that operates over an entire grid like colorizing all rows, we need to undo each cell operation.

tobiolo: I like this idea, let us see how it can be implemented :)

PR #1158, opened 2025-08-30
Configure and normalize line endings

tobiolo:


README.md
```diff
@@ -1,93 +1,93 @@
-<p align="center">
```
aardappel, line 1: why are all files changed? What line endings did they have before and what do they have now?

Your change should only affect line endings locally, and not in the repo, should it not?

PR #1172, opened 2025-09-02
Update RU translation. Based on #1162 updated translation template.

qua-rus:


tobiolo:
Thank you for your translation update :-)

PR #1187, opened 2025-09-09
Rename doc to document in TreeSheetsScriptImpl

tobiolo:


tobiolo:
Thanks for your feedback on this matter. I will take it into account.

PR #1190, opened 2025-09-10
Add generation to undo stack

tobiolo:


src/treesheets_impl.h
```diff
@@ -1,36 +1,19 @@
...
         document = sys->frame->GetCurrentTab()->doc;
         current = document->root;
-        lowestcommonancestor = nullptr;
-    }
-
-    void AddUndoIfNecessary() {
-        if (!lowestcommonancestor) {
-            UpdateLowestCommonAncestor();
-        } else {
-            for (auto p = current; p; p = p->parent) {
```
aardappel, line 18: Actually, this was still useful code.. if a script generates thousands of changes, this code will make that into very few undo points.

tobiolo: Ah yes, you are completely right.

PR #1206, opened 2025-09-22
Upgrade to newer C++ 23 standard

tobiolo:


aardappel:
Cutting edge! Let's see if we get complaints from anyone building on older systems.. :)

PR #1214, opened 2025-09-28
Move auxiliary stuff into helper functions

tobiolo:


src/tsapp.h
```diff
@@ -117,6 +97,14 @@ struct TSApp : wxApp {
         return 0;
     }
 
+    void SetupLanguage() {
```
aardappel, line 100: Unless the code in question is crazy long & complicated, I generally feel that functions that are used exactly once make code readability worse, not better, compared to a block of code (with maybe a comment stating what the block does).

PR #1215, opened 2025-09-28
We already guarantee for wxUSE_UNICODE set to true in TSApp::OnInit()

tobiolo:


src/text.h
```diff
@@ -18,12 +18,7 @@ struct Text {
     }
 
     size_t EstimatedMemoryUse() {
-        return sizeof(Text) + t.Length() *
-            #if wxUSE_UNICODE
-            sizeof(wchar_t);
-            #else
-            sizeof(char);
-            #endif
+        return sizeof(Text) + t.Length() * sizeof(wchar_t);
```
aardappel, line 21: good to at least add an `assert` if this code heavily depends on this setting being correct, given how non-local this setting is.

PR #1280, opened 2025-10-28
Fix MacOpenFiles

tobiolo:
- For virtual override the exact function signature is needed.
- Event loop needs to be manually triggered with MacOpenFiles.


tobiolo:
Now open the TreeSheets file via Finder should work in both cases a new instance is launched or the running one is used.

PR #1285, opened 2025-11-03
Only on wxGTK there is \r\n

tobiolo:


tobiolo:
Yup, that is insane but that's what I debugged on Wayland so far.

PR #1313, opened 2025-11-29
More robust epoch rename

jn64:
Should have done it this way in the first place, my bad for hard-coding the epoch


tobiolo:
Thanks for the improvement :-)

PR #1354, opened 2026-01-01
Use wxAuiToolbar instead of wxToolBar

tobiolo:


georgeraraujo:
What was the rationale for this? It is not as clean as the original toolbar, and takes more screen estate.

georgeraraujo:
The toolbar takes more space because of the drag handles, and the colors of the labels next to the dropdown lists don't match those of the buttons.

Before:
image


After:
image

PR #1362, opened 2026-01-29
Added Note functionality and fix macOS launch issues

fidelnamisi:
Edit > Edit Note
This creates a note that is associated with a cell.


tobiolo:
Thank you very much for your PR!
I allowed myself to pick the note functionality from your PR. I would not rename it to treesheetsv2 but just leave the configuration files as is. I hope this is ok for you.

PR #1374, opened 2026-02-23
Add option to use system wxWidgets libraries

bkmgit:


tobiolo:
https://cmake.org/cmake/help/latest/module/FetchContent.html#:~:text=in%20version%203.24%3A-,FIND_PACKAGE_ARGS,-This%20option%20is

`FetchContent_Declare` with `FIND_PACKAGE_ARGS` already looks for the system package first. Only if does not find it, it will fallback to the download and internal static linking.

PR #1380, opened 2026-03-06
Added AI assistant: fix shortcuts, add multi-tab context, optimize …

fidelnamisi:
Feat: Added AI Side-Panel Chat with Multi-Tab Context & UX Fixes

Now you can do an AI chat with all your Treesheets files right from within TreeSheets.


axb21:
Please do not merge PRs that add AI features to Treesheets. I like this software but I will have to stop using it if LLM/AI related features are added.

tobiolo:
While I appreciate the effort and I also use AI, I would like to keep TreeSheets to be self-contained without adding dependency to online web services. It is still possible to copy stuff into clipboard and then into AI.

aardappel:
I'm curious, how does this work? I presume it is a "bring your own API key" kind of thing? What is sent to the LLM for it to be able to interpret the contents of a TreeSheet? How does the LLM make modifications?

PR #1411, opened 2026-06-04
Add Vietnamese tutorial translation

vuanhvu11982:
Hi! I have added the Vietnamese translation for the tutorial (tutorial-vi.cts).

   Note: Since the default TreeSheets font doesn't render Vietnamese Unicode characters correctly, I highly recommend using the Consolas font for the best display. It works flawlessly and keeps the text perfectly aligned.


tobiolo:
Thank you very much, again @vuanhvu11982-afk :-)!

PR #1413, opened 2026-06-28
Update Korean

VenusGirl:


tobiolo:
Wow @VenusGirl that was fast :) Thank you!

PR #1427, opened 2026-09-06
Update Japanese language files

coolvitto:
Update ts.mo and ts.po


tobiolo:
Thank you @coolvitto !

